PR #4, opened 2018-04-03
feat(internals.js): additional parameters for event emitting functions

theblushingcrow:


src/lib/src/internals.ts
```diff
@@ -91,6 +91,12 @@ export class Spectator<C> {
    * @param selector string | Element | DebugElement
    */
   click(selector: string | Element | DebugElement) {
```
NetanelBasal: We should have a type, instead of repeating this. 
`type SpectatorElement = string | Element | DebugElement | Window`

src/lib/src/internals.ts
```diff
@@ -91,6 +91,12 @@ export class Spectator<C> {
...
+    const element = this.getNativeElement(selector);
+    element.click();
+    this.detectChanges();
+  }
+
+  getNativeElement(selector: string | Element | DebugElement | Window) {
```
NetanelBasal: Add a doc block, please.

PR #10, opened 2018-04-09
Fix `dispatchKeyboardEvent' not working (#9)

### LINKED ISSUE #9: dispatchKeyboardEvent doesn't work as expected
Adding the following to `ZippyComponent`:
```html
<div (click)="toggle()" class="zippy__title" (keyup.enter)="toggle()">
```

and adding the following test:
```typescript
it('should toggle the content when pressing "Enter"', () => {
    host = createHost(`<zippy title="Zippy title"></zippy>`);

    const pressEnter = () => {
      host.dispatchKeyboardEvent('.zippy__title', 'keyup', /* Enter key */ 13);
    };

    pressEnter();
    expect(host.query(".zippy__content")).toExist();

    pressEnter();
    expect(".zippy__content").not.toExist();
  });
```

Will **fail**, though it's expected to **succeed**.

The following alternative code in (plain) Angular **works** (notice the differences in the `pressEnter` function):
```typescript
it('should toggle the content when pressing "Enter"', () => {
    host = createHost(`<zippy title="Zippy title"></zippy>`);

    const pressEnter = () => {
      host.getNativeElement('.zippy__title').dispatchEvent(new KeyboardEvent('keyup', { key: 'Enter' }));
      host.detectChanges();
    };

    pressEnter();
    expect(host.query(".zippy__content")).toExist();

    pressEnter();
    expect(".zippy__content").not.toExist();
  });
```

bengry:
This PR fixes #9 (at least partially)

### DIFF
--- a/src/app/zippy/zippy.component.spec.ts
+++ b/src/app/zippy/zippy.component.spec.ts
@@ -9,7 +9,7 @@ import { fakeAsync, tick } from "@angular/core/testing";
 describe("ZippyComponent", () => {
   let host: SpectatorWithHost<ZippyComponent>;
 
-  const createHost = createHostComponentFactory(ZippyComponent);
+  const createHost = createHostComponentFactory<ZippyComponent>(ZippyComponent);
 
   it("should display the title", () => {
     host = createHost(`<zippy title="Zippy title"></zippy>`);
@@ -68,7 +68,7 @@ describe("ZippyComponent", () => {
     expect(".zippy__content").not.toExist();
   });
 
-  it("should toggle the content", () => {
+  it("should toggle the content when clicked", () => {
     host = createHost(`<zippy title="Zippy title"></zippy>`);
 
     host.click(".zippy__title");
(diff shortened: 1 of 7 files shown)


src/app/zippy/zippy.component.spec.ts
```diff
@@ -77,6 +77,20 @@ describe("ZippyComponent", () => {
...
     expect(".zippy__content").not.toExist();
   });
+
+  it('should toggle the content when pressing "Enter"', () => {
+    host = createHost(`<zippy title="Zippy title"></zippy>`);
+
+    const pressEnter = () => {
+      host.dispatchKeyboardEvent('.zippy__title', 'keyup', 'Enter');
+    };
+
```
NetanelBasal: I think it will be helpful to add the following method:

`host.pressEnter('.zippy__title') // if empty it should be on the tested element `

What do u think? Can you add it, please? ( you can add it here https://github.com/NetanelBasal/spectator/blob/master/src/lib/src/internals.ts#L120)

NetanelBasal: Also, add your details to the contributors [file](https://github.com/NetanelBasal/spectator/blob/master/.all-contributorsrc).

bengry: is Enter that common to warrant such a method? I don't want to over-pollute the API. I mean, if this is added, next thing to be added is Space etc.

bengry:
@NetanelBasal Should I also bump the library version, or are you going to add it directly (among with other changes I assume) as a 1.1.0/1.0.1 version?

NetanelBasal:
Yes, there is no need.

PR #24, opened 2018-06-20
feat(directive): add MockDirective helper function

benelliott:
Thanks for the library, it's a great way of making my unit tests more concise! However I encountered some problems when trying to mock directives. I couldn't mock them with `MockComponent` because they were being applied to an element that already had a component registered on it. So this commit adds a `MockDirective` class that is exactly symmetrical to `MockComponent`.


NetanelBasal:
Thanks. Sounds correct.

PR #26, opened 2018-06-20
feat(mock-component): content projection for default MockComponent

dirkluijk:
This changes the default template for `MockComponent` from `''` to `'<ng-content></ng-content>'`, so that the content of mocked components are projected by default. Usually you want to test the projected contents as well, because they are part of their host component.


NetanelBasal:
I see what you mean, but It could lead to a breaking change. Imagine someone that doesn't want this behavior will get an error for not providing the components used inside the ng-content.

NetanelBasal:
Let's add it in version 2.0 with angular 6.

PR #32, opened 2018-06-26
feat(query): add dom-testing-library query tools

benelliott:
This PR allows you to use selectors based on dom-testing-library in `query`, `queryHost` and `queryAll` methods. An example of the usage is included in the demo app.

```html
<input type="text" placeholder="By placeholder" id="by-placeholder-input">
```

```typescript
it("should allow querying by placeholder", () => {
   const element = spectator.query(byPlaceholder("By placeholder"));
   expect(element).toHaveId("by-placeholder-input");
});
```

Note that this uses `dom-testing-library` v2.1.0 because it is the last version before typings were introduced for TypeScript. These typings use the `infer` keyword which will throw a compile error on anything below TypeScript 2.8, which Angular does not currently support. When Angular is updated to support TS2.8 we can update `dom-testing-library`.

Some changes were made to the method signatures of the `query*` methods but this PR is intended to be backwards-compatible.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,10 +10,10 @@
       "integrity": "sha512-U0BCZtThq5rUfY08shHXpxe8ZhSsiYB/cJjUvAWRTs/ORrs8pbngS6xwseQws8d/vHoVrtqGD9GU9h8AmFRERQ==",
       "dev": true,
       "requires": {
-        "loader-utils": "^1.1.0",
-        "source-map": "^0.5.6",
-        "typescript": "~2.6.2",
-        "webpack-sources": "^1.0.1"
+        "loader-utils": "1.1.0",
+        "source-map": "0.5.7",
+        "typescript": "2.6.2",
+        "webpack-sources": "1.1.0"
       },
       "dependencies": {
         "source-map": {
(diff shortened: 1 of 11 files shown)


benelliott:
I would be happy to add some documentation for this too but not sure where it lives?

NetanelBasal:
Looks great, thanks. I need to add you as a writer to gitbook so you can add the docs. Can you give me an email address so I can send you the invite link, please?

benelliott:
Sure, I'll send an email to the address you've got in the repo.

NetanelBasal:
Sent you a link. Btw you are going with the refactoring to the Angular library? ( if not I will do this )

benelliott:
I have had a go and got most of the way there, but need to rebase and make sure all your tooling still works (precommit hooks etc). I'm busy for a few nights this week though so if it can't wait for another week or so then by all means go ahead.

dirkluijk:
Why has `queryAll<R = Element[]>(): R;` been changed to `queryAll<R extends Element[]>(): R[];`? It now returns a multidimensional array of `Element[][]` by default (which is a breaking change, but it seems unintended).

benelliott:
@dirkluijk That seems to be a typo and should be `queryAll<R extends Element>(): R[];`. Thanks.

PR #58, opened 2018-12-17
Mock provider properties

MartinNuc:
Solves https://github.com/NetanelBasal/spectator/issues/56

First commit shows the problem.

The second commit adds optional argument for `mockProvider` representing a kind of template for a mock. It's supposed to be an object with subset of properties from the mocked service.

### DIFF
--- a/src/app/consumer.service.spec.ts
+++ b/src/app/consumer.service.spec.ts
@@ -6,7 +6,11 @@ import { Subject } from 'rxjs';
 describe('ConsumerService', () => {
   const spectator = createService({
     service: ConsumerService,
-    providers: [mockProvider(ProviderService)]
+    providers: [
+      mockProvider(ProviderService, {
+        obs$: new Subject()
+      })
+    ]
   });
 
   it('should consume mocked service with properties', () => {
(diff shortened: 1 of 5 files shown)


MartinNuc:
@NetanelBasal can you please point me where to update documentation? Couldn't find it in the repository.

src/app/consumer.service.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+import { createService, mockProvider } from '@netbasal/spectator';
+import { ProviderService } from './provider.service';
+import { Subject } from 'rxjs';
+
+describe('ConsumerService', () => {
+  const spectator = createService({
+    service: ConsumerService,
+    providers: [
+      mockProvider(ProviderService, {
+        obs$: new Subject()
```
NetanelBasal: I'm wondering what is the point here as it's already a Subject. That's a dummy example for how to use?

NetanelBasal: Forget about it :)

NetanelBasal:
Thanks, I need to add you to our gitbook so you can edit the docs. To where I can send you an invitation?

MartinNuc:
martin@nuc.cz

Thanks

PR #61, opened 2018-12-23
test(jest): add Jest specs matching all the Jasmine specs that have Spectator imports

LayZeeDK:
Add Jest specs to exercise as much of the API as the Jasmine specs do.

I discovered that the DOM selectors have not been exposed in Spectator's Jest API.


NetanelBasal:
Thanks!

PR #70, opened 2019-01-19
Fix for using query in directives

MartinNuc:
In previous [PR](url
https://github.com/NetanelBasal/spectator/pull/69) I made fix in incorrect class.

This reverts previous change and provides fix for in correct class.

Also fixes `detectChanges` in the `Spectator` class. Hopefully wont break anything. If it does we need test for that :-)


projects/spectator/src/lib/internals.ts
```diff
@@ -54,7 +47,7 @@ export class Spectator<C> {
    * Run detect changes on the host component
```
MartinNuc: hmm I should also update the comment here.

@NetanelBasal do you remember why it was done this way? Is there a time when we need to run change detection for host when not having the host?

MartinNuc: ✅ comment fixed.

NetanelBasal:
I'm reverting the previous PR for now and will go over this PR when I have time. I don't want to rush.

MartinNuc:
@NetanelBasal I reverted the previous PR as part of this one. I mean the commit which was wrong. The second commit were just failing tests to point out the issue. If you revert the previous PR I will rebase this to remove revert commit.

Please take extra look on the change of [detectChanges](https://github.com/NetanelBasal/spectator/pull/70/files#diff-fef2a17cab358daf4e30aebe3406a55bL57) in Spectator class.

Sure take your time. And no need to release a new version after merging. We dont depend on it.

PR #76, opened 2019-02-17
feat(config): add declareComponent option

oschlegel:
I suggest to introduce a `declareComponent` option to the SpectatorOptions, usecase would be if the component i want to test is already declared in a module together with it's dependecies. This benefits the reusability of already existing modules and simplifies setting up tests for components. Example:

```
@NgModule({
  declarations:[TheComponent, SubComponent],
  exports:[TheComponent],
  imports: [SomeModule, AnotherModule]
})
export class TheComponentModule {}
```

With `declareComponent` option:

```
const createHost = createHostComponentFactory({
    component: TheComponent,
    imports: [TheComponentModule],
    declareComponent: false
});
```

Without `declareComponent` option:

```
const createHost = createHostComponentFactory({
    component: TheComponent,
    declarations:[SubComponent],
    imports: [SomeModule, AnotherModule]
});
```

### DIFF
--- a/projects/spectator/src/lib/config.ts
+++ b/projects/spectator/src/lib/config.ts
@@ -26,6 +26,7 @@ export type SpectatorOptions<T = any, H = HostComponent> = TestModuleMetadata & 
   componentProviders?: any[];
   mocks?: Type<any>[];
   detectChanges?: boolean;
+  declareComponent?: boolean;
 };
 
 const defaultOptions: SpectatorOptions<any, any> = {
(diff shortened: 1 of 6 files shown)


MitkoTschimev:
You mean if you want to test the component as an integration test? Than it make sense. I know this feature from the angular playground and there it saves really a lot of time!!! So I would like to see this improvement in spectator!!

NetanelBasal:
Thanks, please add dummy tests (with host and without a host) that test that it works as expected.

Issue #78, opened 2019-02-27
Incorrect keyboard event creation


jpzwarte:
So when i use the spectator API to dispatch a keydown event like this:

```
spectator.dispatchKeyboardEvent(select, 'keydown', SPACE);
```

Things do not work correctly. After debugging the code, it turns out that `event.metaKey` is true for the dispatched event (running in Chrome). After some looking, i think the bug is here: https://github.com/NetanelBasal/spectator/blob/master/projects/spectator/src/lib/event-objects.ts#L40

When i look at MDN, the API for `initKeyEvent` is different from `initKeyboardEvent`. The spectator code doesn't seem to take that into account. Also, according to MDN both methods are deprecrated and you should just use the `new KeyboardEvent(...)` constructor function.

```
> ev = document.createEvent('KeyboardEvent')
KeyboardEvent {isTrusted: false, key: "", code: "", location: 0, ctrlKey: false, …}
> ev.initKeyboardEvent('keydown', true, true, window, 0, 0, 0, 0, 0, 13)
undefined
> ev.metaKey
true
```

NetanelBasal:
Actually, the code is from material. https://github.com/NetanelBasal/spectator/blob/master/projects/spectator/src/lib/event-objects.ts#L1

NetanelBasal:
They updated the code: https://github.com/angular/material2/commit/04716b1ad0da60c52618e8ab8072bdee5c2b5eb8#diff-27d3bcc8e3ad6d38c08edce8bef3b9c7

Thanks.

PR #79, opened 2019-02-27
Fix keyboard event creation.

jpzwarte:
This attempts to fix issue #78. I don't see any unit tests that i can update to make sure the fix is correct?


jpzwarte:
I used the same mechanism as material uses: https://github.com/angular/material2/blob/master/src/cdk/testing/event-objects.ts#L62

PR #82, opened 2019-03-08
Add schematics

rexebin:
Add schematics for generating component, service and directive with Spectator specs.
The PR would add the following schematics:
`ng g spectator-component spec-comp`
`ng g spectator-component spec-comp-host --withHost=true`
`ng g spectator-component spec-comp-custom-host --withCustomHost=true`
`ng g spectator-directive spec-directive`
`ng g spectator-service spec-service`
`ng g spectator-service spec-data-service --isDataService=true`

Tested in a Nx workspace.


NetanelBasal:
Wow, that's was quick. Looks great. I have two questions:
1. There is no way to use the default CLI schematics and just replace the spec file?
2. If not, we should make the command shorter or create an alias. Something like:
```
ng g cs spec-comp
ng g cs spec-comp-host --withHost=true
ng g cs spec-comp-custom-host --withCustomHost=true
ng g ds spec-directive
ng g dse spec-service
ng g dse spec-data-service --dataService=true
```

rexebin:
1. At the moment, the schematics execute default cli and put a Spectator spec file in place, I do not know a way to use the cli command and automatically replace the spec
2. it can support alias. For service, it should be `ng g ss some-service`, basically add a s after or before the default alias would be easy to remember.

NetanelBasal:
Great, so add the alias option, please. A couple of things I notice:
1. You need to update the README file with the commands.
2. You need to add yourself as a contributor, check the command in `package.json`.
3. You need to exclude to the folder from Jest. The tests are failing now. If you need help with that lmk. 

Thanks!

rexebin:
Aliases added.
I will try to do the rest.
I am not sure if the schematic would work just like this after installing the lib. It works in my nx workspace but it has a different setup. We will need to look into it if it doesn't work.

NetanelBasal:
I will test it before release. You should also extend the default CLI. https://github.com/datorama/akita-schematics/blob/master/src/collection.json#L2

rexebin:
@NetanelBasal I do need help for excluding the folder from Jest. Please help. 
Thanks.

NetanelBasal:
I fixed the tests. I'm testing the code, and there is a problem. The test file isn't in the same folder as the component files.
image


You can test it by running:
`
ng add @netbasal/spectator@3.5.0-beta.0
ng config cli.defaultCollection @netbasal/spectator
ng g cs tabs
`

rexebin:
I launch the schematic with webstorm build in tools, it gave a path. Looks
like using cli on command line don’t do the same. I will try to fix it.

Netanel Basal <notifications@github.com>于2019年3月10日 周日上午7:41写道：

rexebin:
@NetanelBasal I am sorry about the bug. I added a function to ensure the path is valid, essentially the same with Angular Cli. Please try it out. It should put the test file into the same folder.

PR #86, opened 2019-03-12
Prevent error when component is not found By.directive

MartinNuc:
Solves https://github.com/NetanelBasal/spectator/issues/85

### DIFF
--- a/projects/spectator/src/lib/errors.ts
+++ b/projects/spectator/src/lib/errors.ts
@@ -1,1 +1,14 @@
-export class SpectatorError extends Error {}
+export class SpectatorError extends Error {
+  constructor(message: string) {
+    super(message);
+    // https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
+    Object.setPrototypeOf(this, SpectatorError.prototype);
+  }
+}
+export class SpectatorDebugElementNotFoundError extends SpectatorError {
+  constructor(message: string) {
+    super(message);
+    // https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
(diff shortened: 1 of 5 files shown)


projects/spectator/src/lib/errors.ts
```diff
@@ -1 +1,14 @@
-export class SpectatorError extends Error {}
+export class SpectatorError extends Error {
+  constructor(message: string) {
+    super(message);
+    // https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
+    Object.setPrototypeOf(this, SpectatorError.prototype);
+  }
+}
+export class SpectatorDebugElementNotFoundError extends SpectatorError {
```
MartinNuc: I created a specific error class for this case.

src/app/zippy/zippy.component.ts
```diff
@@ -38,6 +33,11 @@ export class ZippyComponent {
   visible = false;
   updatedAsync = false;
 
+  @HostListener('keyup.esc')
+  onEsc() {
+    this.toggle();
+  }
```
MartinNuc: linter complained here that method definitions are supposed to be after properties.

projects/spectator/src/lib/internals.ts
```diff
@@ -70,7 +70,15 @@ export class Spectator<C> {
...
-    return _getChild<R>(this.debugElement)(directiveOrSelector, options);
+    try {
+      return _getChild<R>(this.debugElement)(directiveOrSelector, options);
+    } catch (err) {
+      if (err instanceof SpectatorDebugElementNotFoundError) {
+        return null;
+      } else {
+        throw err;
+      }
+    }
```
NetanelBasal: What about queryAll?

MartinNuc: Tested that one and it worked fine so no change is required.

Test for it is here: https://github.com/NetanelBasal/spectator/pull/86/files#diff-e6572426b83a2437724cd9a71d19a7deR92

PR #94, opened 2019-03-13
Stop mockProvider to modify its input

MartinNuc:
Fixes https://github.com/NetanelBasal/spectator/issues/93

### DIFF
--- a/src/app/mocks.spec.ts
+++ b/src/app/mocks.spec.ts
@@ -1,0 +1,15 @@
+import { mockProvider } from '@netbasal/spectator';
+import { WidgetService } from './widget.service';
+
+describe('mockProvider', () => {
+  it('should not modify the object passed in 2nd argument when running the mock factory', () => {
+    const customPropertiesAndMethods: Partial<Record<keyof WidgetService, any>> = {
+      testingProperty: 'overriden'
+    };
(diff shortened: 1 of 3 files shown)


src/app/mocks.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+    const customPropertiesAndMethods: Partial<Record<keyof WidgetService, any>> = {
+      testingProperty: 'overriden'
+    };
+    const { useFactory: factory } = mockProvider(WidgetService, customPropertiesAndMethods);
+    factory();
+    expect(customPropertiesAndMethods).toEqual({
+      testingProperty: 'overriden'
+    });
+  });
+});
```
MartinNuc: This is really a unit test for `mockProvider` function. Imho should be located somewhere close to `mock.ts` but there is no place for unit tests. I placed it here but it should be probably moved in the future.

eweap:
@MartinNuc Hey thanks for the PR !

@NetanelBasal @MartinNuc Should this fix also be applied to the jest implementation of `createSpyObject` ?

MartinNuc:
Unfortunately I am not sure about `createSpyObject` :-( Maybe we should create test for jest and see how it behaves.

PR #95, opened 2019-03-14
run ngOnDestroy for service after every test

MartinNuc:
Fixes https://github.com/NetanelBasal/spectator/issues/92

I tried to simulate the tests exactly like it happened to us. Part of the problem is also that a single subject is used across multiple tests (could be issue with `ReplaySubject` for example). If it created a new subject every time it would cause only memory leak :-)

Also this sort of black-box-testing depends on the order of tests. General approach is to randomise test execution order. Then this test would not be reliable.

### DIFF
--- a/projects/spectator/src/lib/service.ts
+++ b/projects/spectator/src/lib/service.ts
@@ -42,6 +42,13 @@ export function createService<S>(options: Params<S> | Type<S>): SpectatorService
     TestBed.configureTestingModule(module);
   });
 
+  afterEach(() => {
+    const testedService = TestBed.get(service);
+    if (typeof testedService.ngOnDestroy === 'function') {
+      testedService.ngOnDestroy();
+    }
+  });
+
   return {
     get service(): S {
       return TestBed.get(service);
--- a/src/app/ng-on-destroy.service.spec.ts
+++ b/src/app/ng-on-destroy.service.spec.ts
@@ -1,0 +1,36 @@
+import { NgOnDestroyService, ngOnDestroySpy } from './ng-on-destroy.service';
+import { SubjectService } from './subject.service';
+import { createService } from '@netbasal/spectator';
+import { mockProvider } from '@netbasal/spectator/jest';
+import { Subject } from 'rxjs';
+
+describe('NgOnDestroyService', () => {
+  const spyAcrossMultipleTests = jasmine.createSpy('ngOnDestroy test spy');
+  const subjectAcrossMultipleTests$ = new Subject();
+  const spectator = createService({
+    service: NgOnDestroyService,
+    providers: [
+      {
+        provide: ngOnDestroySpy,
+        useValue: spyAcrossMultipleTests
+      },
+      mockProvider(SubjectService, {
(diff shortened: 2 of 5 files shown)


NetanelBasal:
What about creating a new service instance `beforeEach`? It will solve the issue, no?

MartinNuc:
I think a new service instance IS created in `beforeEach` by `TestBed.configureTestingModule(module);`. If you mean calling `mockProvider` again that would still cause memory leak because the rxjs subscription from previous test is never unsubscribed :-(

That happens btw also when someone doesn't implement ngOnDestroy for clean up.

And one more thing - will this solution work with https://github.com/NetanelBasal/ngx-take-until-destroy ?

NetanelBasal:
Yes, it should work. What about passing a factory function someway?

NetanelBasal:
No, it will not work. `mockProvider` function is called once, therefore we always have one instance of `Subject`. I can't see any soultion for that other than using spies and return new `Subject` in every `it`.

NetanelBasal:
If the user passes somehow a factory function, we could do something like:

```ts
  beforeEach(() => {
    (options.providers || []).forEach(type => module.providers.push(type));
    TestBed.configureTestingModule(module);
  });
```

MartinNuc:
The proposed solution still doesn't solve the memory leak from ongoing subscription.

I think the main problem is that the same subject is used across multiple tests:

```
mockProvider(SubjectService, {
   subject$: subjectAcrossMultipleTests$
})
```

We need to provide a new subject for every test. Therefore as you say - use factory method for mocking properties:

```
let newSubjectForEveryTest$;

...

mockProvider(SubjectService, {
   subject$: () => { // "factory" - but very confusing API ... I am againts this one.
      newSubjectForEveryTest$ = new Subject();
      return newSubjectForEveryTest$;
   }
})

...

beforeEach(() => {
   newSubjectForEveryTest$ = spectator.get(SubjectService).subject$;
})
```

dirkluijk:
Interesting topic.

Just as a side note, apart from finding a solution, isn't it a bad practice to subscribe from within a service? I mean, in the end your components are always somehow the consumers? If yes, then I would provide a transformed observable in the service. And components life-cycles are always handled automatically.

MartinNuc:
@dirkluijk in general it doesn't have to be a subscription. It could be just `setInterval` and it creates same problem.

Dunno if it's a bad practice. We have a service which listens for router navigation events. It's not related to any component at all.

dirkluijk:
Valid examples. 😉 I will try to help finding a nice solution.

dirkluijk:
@NetanelBasal can we merge this PR?

NetanelBasal:
@MartinNuc I can't remember. Does this solution work?

MartinNuc:
@NetanelBasal yes it works although it doesn't solve all cases. It's more about the way how one writes tests.

Nevertheless I think ngOnDestroy should be called anyways - that's what this PR brings.

dirkluijk:
Closed in favour of https://github.com/ngneat/spectator/pull/161.

PR #96, opened 2019-03-20
fix(jest): fix for incorrect typing of SpyObject for Jest

dirkluijk:
Just upgraded to Jest 24, and got compiler errors. I think the reason this bug was nog blocking before, is that `@types/jest@24.0.0` is more strict with generics.

**The issue**
Given an interface:
```typescript
interface Person {
  name: string;
  saySomething: () => string;
}
```

when you define `SpyObject<Person>` as `{ [P in keyof Person]: Person[P] & jest.Mock<Person>`, then `personSpyObject.saySomething` would be a `() => string & jest.Mock<Person>`. That is wrong, because it should be a `jest.Mock<string>`.

To get the return type of `Person[P]` I use `ReturnType` combined with conditional types from Typescript 2.8. Awesome stuff by the way.

### DIFF
--- a/projects/spectator/jest/src/mock.ts
+++ b/projects/spectator/jest/src/mock.ts
@@ -1,7 +1,7 @@
 import { Provider, Type } from '@angular/core';
 import { CompatibleSpy, SpyObject as BaseSpyObject, installProtoMethods } from '@netbasal/spectator';
 
-export type SpyObject<T> = BaseSpyObject<T> & { [P in keyof T]: T[P] & jest.Mock<T> };
+export type SpyObject<T> = BaseSpyObject<T> & { [P in keyof T]: T[P] & (T[P] extends (...args: any[]) => infer R ? jest.Mock<R> : T[P]) };
 
 export function createSpyObject<T>(type: Type<T>, template?: Partial<Record<keyof T, any>>): SpyObject<T> {
   const mock: any = template || {};
--- a/projects/spectator/package.json
+++ b/projects/spectator/package.json
@@ -27,5 +27,8 @@
     "jquery": "3.3.1",
     "schematics-utilities": "^1.1.1"
   },
+  "peerDependencies": {
+    "typescript": ">= 2.8.0"
+  },
   "schematics": "./schematics/collection.json"
 }


dirkluijk:
Build is now stable.

NetanelBasal:
It might be a breaking change because maybe there are people that from some reason are still in < 2.8.

dirkluijk:
Strictly speaking yes.

I checked it on our project, but this change seems to be perfectly backwards compatible with Jest 23.x.

The reason is that it was already wrong since the beginning of the Jest support in Spectator, but Jest 24 has a lot of TypeScript improvements and on Jest 23.x one would not get any errors from it.

dirkluijk:
Oh, sorry. You meant Typescript < 2.8.
Yes, totally right. Didn't think of that.

Angular 6 depends on TypeScript 2.7
but Angular 7 depends on TypeScript 3.1
and Angular 8 depends on TypeScript 3.3.

Can we drop support for Angular 6 and introduce a new (major?) version of Spectator?

NetanelBasal:
Yes, and you could also upgrade to Jest 24?

dirkluijk:
Yes, but this fix is required for Jest 24. ;) Otherwise you get compile errors.

NetanelBasal:
Sure

dirkluijk:
@NetanelBasal what fix would you suggest for now? `Mock<any>` or the breaking TS 2.8 change? 😁

NetanelBasal:
I'm suggesting to release a new major version and update both Jest and Typescript.

dirkluijk:
Just simplified it a bit and added Typescript 2.8 as peer dependency.

NetanelBasal:
Does this require a major release?

dirkluijk:
I ‘m not sure. Spectator API itself hasn’t changed, but its peer dependencies have.

Strictly speaking, it should. I know Angular does TS upgrades for major versions as well.

See also https://github.com/semver/semver/issues/148.

It’s up to you to decide. ;)

NetanelBasal:
I'm not sure if someone is still using Angular 6 🤔. I don't think it's worth a new major version.

NetanelBasal:
3.6.0

PR #103, opened 2019-04-18
Add exact text matcher support

YoeriNijs:
During our daily work, we discovered that the toHaveText matcher is not accurate for exact matches. For instance, you can have this line of code:

`expect('foo').toHaveText('bar'); // true`

This is true, of course. However, this is true as well:

`expect('foo').toHaveText(''); // true`
`expect('foo').toHaveText('ar'); // true`

In order to have an exact text matcher, I have added the toHaveExactText matcher. It works as follows:

`expect('foo').toHaveExactText('bar'); // true`
`expect('foo').toHaveExactText('ar'); // false`

Furthermore, one can call the exact matcher by adding an additional property to the original matcher as well:

`expect('foo').toHaveText('ar', false); // true`
`expect('foo').toHaveText('ar', true); // false`

What do you guys think of this? Please, feel free to change the code in this PR if you believe your change fits better.

### DIFF
--- a/projects/spectator/jest/src/matchers-types.d.ts
+++ b/projects/spectator/jest/src/matchers-types.d.ts
@@ -20,7 +20,9 @@ declare namespace jest {
 
     toHaveProperty(prop: string, val: string | boolean): boolean;
 
-    toHaveText(text: string | Function): boolean;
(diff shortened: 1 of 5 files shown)


NetanelBasal:
Thanks!

PR #113, opened 2019-06-11
fix(matchers): visibility matchers work with Jest

### LINKED ISSUE #97: feat: make toBeVisible take into account 'hidden' attribute
Hey folks,
I noticed `toBeVisible` matcher ignore [`hidden` attribute](https://www.w3schools.com/tags/att_global_hidden.asp).
It may be a good idea to check that too.

IlCallo:
I recreated the visibility check to be jsdom friendly to make them work with Jest too.
Mixes jQuery 1 and 3 :visible and :hidden pseudo-selector implementation, because jQuery 3 implementation won't work in jsdom: layout isn't managed.

Also checks for 'hidden' attribute.

(I accidentally closed previous PR while force pushing an update)

Fixes #97

### DIFF
--- a/projects/spectator/src/lib/matchers.ts
+++ b/projects/spectator/src/lib/matchers.ts
@@ -240,18 +240,42 @@ export const toBeEmpty = comparator(el => {
 });
 
 /**
- * The :hidden selector selects hidden elements.
- * Hidden elements are elements that are:
- * 1. Set to display:none
+ * Hidden elements are elements that have:
+ * 1. Display property set to "none"
  * 2. Width and height set to 0
  * 3. A hidden parent element (this also hides child elements)
- * 4. Form elements with type="hidden"
+ * 4. Type equal to "hidden" (only for form elements)
+ * 5. A "hidden" attribute
+ */
+function isHidden(el) {
+  while (el) {
+    if (el === document) {
+      break;
+    }
+
+    if (!(el.offsetWidth || el.offsetHeight || el.getClientRects().length) || el.style.display === 'none' || el.style.visibility === 'hidden' || el.type === 'hidden' || el.hasAttribute('hidden')) {
+      return true;
+    }
+
+    el = el.parentNode;
+  }
+
+  return false;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


projects/spectator/src/lib/matchers.ts
```diff
@@ -240,18 +240,42 @@ export const toBeEmpty = comparator(el => {
...
+ * Hidden elements are elements that have:
+ * 1. Display property set to "none"
  * 2. Width and height set to 0
  * 3. A hidden parent element (this also hides child elements)
- * 4. Form elements with type="hidden"
+ * 4. Type equal to "hidden" (only for form elements)
+ * 5. A "hidden" attribute
+ */
+function isHidden(el) {
+  while (el) {
```
NetanelBasal: Why is the check not only on the element?

IlCallo: You mean, why did I abstracted the isHidden function instead of writing it directly inside the matcher function?
Because it's used both for `toBeVisible` and `toBeHidden` matcher.

Otherwise I didn't understood the question, could you rephrase it?

NetanelBasal: I mean, why traversed the tree? The check should be only on the element itself.

benelliott: @NetanelBasal In the old JSDoc it says

```
 * Hidden elements are elements that are:
...
 * 3. A hidden parent element (this also hides child elements)
```

Which is probably taken from https://api.jquery.com/hidden-selector/

So the previous jQuery implementation was probably using this methodology too.

NetanelBasal: Yes, I know. I meant, do we really need it in a spec context? Usually I want to test the exact element.

benelliott: Dunno, but changing it will probably need a major version bump.

NetanelBasal: Yes, let's leave it like this.

PR #123, opened 2019-06-24
fix(matchers): pass input to `isHidden` through jQuery (fixes #122)

### LINKED ISSUE #122: Bug: `el.getClientRects is not a function` on version 3.9.2
Hi,

I've just upgraded to 3.9.2, and I'm seeing this exception when using `isVisible()` matcher:

`el.getClientRects is not a function`.

Downgrading to 3.9.1 resolves the issue.
Using Headless Chrome (latest build), with Jasmine and Karma.

benelliott:
Ensures the matcher logic operates on an element even if a selector is provided as input to the function.

### DIFF
--- a/projects/spectator/src/lib/matchers.ts
+++ b/projects/spectator/src/lib/matchers.ts
@@ -247,7 +247,9 @@ export const toBeEmpty = comparator(el => {
  * 4. Type equal to "hidden" (only for form elements)
  * 5. A "hidden" attribute
  */
-function isHidden(el) {
+function isHidden(elOrSelector) {
+  let el = $(elOrSelector);
+
   while (el) {
     if (el === document) {
       break;


projects/spectator/src/lib/matchers.ts
```diff
@@ -247,7 +247,9 @@ export const toBeEmpty = comparator(el => {
  * 4. Type equal to "hidden" (only for form elements)
  * 5. A "hidden" attribute
  */
-function isHidden(el) {
+function isHidden(elOrSelector) {
+  let el = $(elOrSelector);
```
NetanelBasal: If I remember correctly, this will not work as JQuery doesn't return the native element. We should obtain the first native element: 

```ts
let el = $(elOrSelector)[0]
if(!el) return undefined
```ts

benelliott: Yeah ;) hence closing the PR!

NetanelBasal: Oh didn't notice :)

Issue #131, opened 2019-07-25
[jest] it.each type conflict


GregOnNet:
Hi @NetanelBasal,

thank you for providing `spectator` I am a big fan of the API you have created.
I tried your library and it seems that some type-definitions of `spectator` still override jest's type definitions.

Importing `@netbasal/spector/jest` causes a compiler error when `it.each` is used.

I prepared a repository reproducing this behaviour: 

- git clone https://github.com/GregOnNet/spectator-jest-playground
- cd spectator-jest-playground
- yarn
- visit `src/app/app.component.spec.ts`

Please let me know if I can provide further information

Thanks in advance
Gregor

NetanelBasal:
@dirkluijk thoughts?

GregOnNet:
I missed to mention, that `test.each` works fine.
Only `it.each` causes problems.

PR #132, opened 2019-07-25
[jest] allow usage of it.each

GregOnNet:
Hi this PR should resolve issue #131.

- Basically, I enforced the usage of jest's type definitions by adding a type reference.
- I added a test to the angular application using `it.each`
- In order to get the test passing I needed to update `@angular-builders/jest` and `jest`
  - There was an issue with `it.each` in an older version of `jest-preset-angular` (see: https://github.com/facebook/jest/issues/6612)

I hope this helps.
Please let me know if this PR needs any changes.


NetanelBasal:
Thanks. Are you sure this isn't a breaking change?

GregOnNet:
Yet, I am not able to tell this, since I am about to start with spectator.
I have no reference project where I can test it.

Do you have another playground repo that I could use to test the introduced changes?

GregOnNet:
Maybe I was asking the wrong question.
I added a test to the Angular app living in this repository, without changing other tests or the implementation.

This made me confident to submit the PR?
What else can I check to ensure that no breaking change has been introduced?

Thanks in advance
Greg

NetanelBasal:
You changed Jest from v23 to v24 which is a major version. We need to check if there is a breaking change there.

NetanelBasal:
@dirkluijk what do you think? I'm not working with Jest, so it's hard for me to say.

dirkluijk:
I don't understand why this fixes https://github.com/NetanelBasal/spectator/issues/131.

I looked into the issue, and I was able to reproduce the error. As soon as you import something from `@netbasal/spectator` or `@netbasal/spectator/jest`, the inferred type ot global `it` changes mysteriously.

That is strange, because in https://github.com/NetanelBasal/spectator/pull/64 I made sure that Spectator does NOT include `jest` neither `jasmine`, so that they can be chosen as peer dependency.

dirkluijk:
OK, I tested this PR on both Jest/Jasmine projects and it seems to raise no issues.

At first sight, I expected issues on Jasmine projects because this PR includes Jest types, but I see they are only included if you reference the secondary entry point `@netbasal/spectator/jest` (which is not the case for Jasmine projects).

So this PR looks good to me!

PR #135, opened 2019-08-02
fix(spectator): improve type inference even more

dirkluijk:
This PR is an improvement of https://github.com/NetanelBasal/spectator/pull/133.

What has been changed:
* Use `Token<T>` instead of `Type<T>`, which allows to use abstract classes as DI token (thanks to @dscheerens). It still allows you to specify a more specific generic type. Added unit tests for this.
* Improve type inference for `SpectatorWithHost.queryHost` and `SpectatorWithHost.queryHostAll` as well

### DIFF
--- a/projects/spectator/jest/src/host.ts
+++ b/projects/spectator/jest/src/host.ts
@@ -6,8 +6,8 @@
  * found in the LICENSE file at https://github.com/NetanelBasal/spectator/blob/master/LICENSE
  */
 
-import { InjectionToken, Type } from '@angular/core';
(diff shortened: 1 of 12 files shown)


dirkluijk:
✔️ Build is now stable

NetanelBasal:
Awesome work, man. Can you confirm that it works both on Jest and Jasmine without breaking change?

dirkluijk:
Confirmed 😄

NetanelBasal:
Something has changed. When building the library, I get the error:

> Entry point @netbasal/spectator has a circular dependency on itself.

dirkluijk:
I 'll take a look.

dirkluijk:
https://github.com/NetanelBasal/spectator/pull/138

PR #145, opened 2019-08-07
feat(mock): allow spy objects and mock providers for abstract classes

dirkluijk:


dirkluijk:
Never mind, this was a bad idea. Abstract classes to exist at runtime, but not their abstract properties so it makes no sense to create a spy object based on their interface. 😄

NetanelBasal:
@dirkluijk could you re-check this file (injection-tokens), please? it fails:

image

Issue #110, opened 2019-05-16
[question]: Angular Routing Testing support


morenoisidro:
Hi, is it planned to support angular routing testing in the near future? Thx!

NetanelBasal:
Sure

LayZeeDK:
Any ideas for what could be useful, @morenoisidro?

Any ideas from you, @NetanelBasal?

NetanelBasal:
We have to check out the "painful" points here.

LayZeeDK:
Exactly

LayZeeDK:
I remembered one thing from testing Router-related things.

If we have a component that subscribes to route parameters, it would be nice to have a utility function to create fake route parameters for testing purposes.

LayZeeDK:
# Testing a routing component
_[Examle from Angular.io](https://angular.io/guide/testing#routing-component)_

```
// dashboard.component.ts
// (...)
export class DashboardComponent {
  constructor(
    private router: Router,
    private heroService: HeroService) {
  }

  gotoDetail(hero: Hero) {
    let url = `/heroes/${hero.id}`;
    this.router.navigateByUrl(url);
  }
}
```

```
// dashboard.component.spec.ts

const routerSpy = jasmine.createSpyObj('Router', ['navigateByUrl']);
const heroServiceSpy = jasmine.createSpyObj('HeroService', ['getHeroes']);

TestBed.configureTestingModule({
  providers: [
    { provide: HeroService, useValue: heroServiceSpy },
    { provide: Router,      useValue: routerSpy }
  ]
});

it('should tell ROUTER to navigate when hero clicked', () => {
  heroClick(); // trigger click on first inner <div class="hero">

  // args passed to router.navigateByUrl() spy
  const spy = router.navigateByUrl as jasmine.Spy;
  const navArgs = spy.calls.first().args[0];

  // expecting to navigate to id of the component's first hero
  const id = comp.heroes[0].id;
  expect(navArgs).toBe('/heroes/' + id,
    'should nav to HeroDetail for first hero');
});
```

Here, we simply want to verify that the side effect of triggering navigation happens when the button is clicked.

It would be nice to be able to add a `Router` spy object spying on methods such as `navigateByUrl` and `navigate`.

LayZeeDK:
# Nested component tests
_[Examples from Angular.io](https://angular.io/guide/testing#nested-component-tests)_

## Stubbing `RouterOutlet`
```typescript
[...]

PR #147, opened 2019-08-07
feat(lib): v4 changes

NetanelBasal:
- API changes discuss in this issue (service factories and new factory options)
- New unit testing suite with coverage thresholds (consider to delete the whole specs files and create a stronger suite from scratch)
- Take a fresh look at the API what to keep, what to add and what to remove/deprecate
- new Wiki? Or updating the existing docs
- A new RoutedSpectator (which extends Spectator like SpectatorWithHost) with some additional routing features (let’s keep this one as last and discuss about it in #110)
- https://github.com/NetanelBasal/spectator/pull/95


src/app/no-overwritten-providers/no-overwritten-providers.component.spec.ts
```diff
@@ -1,10 +1,12 @@
-import { createHostComponentFactory } from '@netbasal/spectator';
+import { createHostComponentFactory, mockProvider } from '@netbasal/spectator';
 import { ComponentWithoutOverwrittenProvidersComponent } from './no-overwritten-providers.component';
+import { DummyService } from './dummy.service';
 
 describe('ComponentWithoutOverwrittenProvidersComponent', () => {
   describe('with options', () => {
-    let createHost = createHostComponentFactory({
-      component: ComponentWithoutOverwrittenProvidersComponent
+    const createHost = createHostComponentFactory({
```
NetanelBasal: I'm not sure how this spec works before. There is a provider on the component's providers.

NetanelBasal:
@dirkluijk you can continue the work on this branch.

src/app/async/async.component.jest.ts
```diff
@@ -21,11 +21,11 @@ describe('ZippyComponent', () => {
...
   });
 
-  it('should be truthy', () => {
-    const host = createHost(`<app-async></app-async>`, false);
-    let queryService = host.get(QueryService);
-    queryService.select.mockReturnValue(of(true));
-    host.detectChanges();
-    expect(host.query('p')).toExist();
-  });
+  // it('should be truthy', () => {
```
NetanelBasal: @dirkluijk doesn't work

dirkluijk:
Hold my beer. 🍺 😄

src/app/auth.service.jest.ts
```diff
@@ -3,19 +3,22 @@ import { DateService } from './date.service';
...
   const spectator = createService({
     service: AuthService,
     mocks: [DateService]
   });
-  it('should not be logged in', () => {
-    const dateService = spectator.get(DateService);
-    dateService.isExpired.mockReturnValue(true);
-    expect(spectator.service.isLoggedIn()).toBeFalsy();
-  });
+  // it('should not be logged in', () => {
```
NetanelBasal: @dirkluijk doesn't work

src/app/injection-tokens.jest.ts
```diff
@@ -1,153 +1,156 @@
...
-import { AbstractQueryService, QueryService } from './query.service';
-import { InjectionToken } from '@angular/core';
-import { ZippyComponent } from './zippy/zippy.component';
-import { WidgetService } from './widget.service';
-
-const MY_TOKEN = new InjectionToken<AbstractQueryService>('some-token');
-
-describe('Injection tokens', () => {
-  describe('with Spectator', () => {
-    const createComponent = createTestComponentFactory({
```
NetanelBasal: @dirkluijk doesn't work

src/app/injection-tokens.spec.ts
```diff
@@ -36,15 +36,15 @@ describe('Injection tokens', () => {
...
-    it('should get by abstract class as token', () => {
-      const service = spectator.get(AbstractQueryService);
-      service.select(); // should compile
-
-      const service2 = spectator.get<QueryService>(AbstractQueryService);
-      service2.selectName(); // should compile
-
-      expect(service instanceof QueryService).toBe(true);
-    });
+    // it('should get by abstract class as token', () => {
```
NetanelBasal: @dirkluijk doesn't work

dirkluijk:
I see why it does not work. It has to do with these changes:

From:
```typescript
get<T>(type: Token<T> | Token<any>, fromComponentInjector = false): T & SpyObject<T> {
```

To:
```typescript
get<T = any>(type: Token<T>, fromComponentInjector = false): T & SpyObject<T> {
```

The latter is not exactly the same as the first one, and maybe (?) not even correct.

I mean, it would be perfectly possible to query for a super class and receive and instance of a sub class:

```
{
  provide: SomeSuperClass,
  useClass: SomeSubClass
}
```

The question is, should it be valid to query for `SomeSubClass` using the token `SomeSuperClass`. If yes, maybe this should be the correct signature:

```typescript
get<R extends T, T = any>(type: Token<T>, fromComponentInjector = false): R & SpyObject<R> {
```

Where R is the result you are expecting, and T the type of the token. In that case:

```
spectator.get(AbstractQueryService)
// infers to:
spectator.get<AbstractQueryService, AbstractQueryService>(AbstractQueryService);

// but you are still able to:
const s: spectator.get<QueryService>(AbstractQueryService);
 // where typeof s = QueryService

// or even:
const s: QueryService = spectator.get(AbstractQueryService);
```

src/app/injection-tokens.spec.ts
dirkluijk: Should be fixed now!

src/app/injection-tokens.jest.ts
dirkluijk: Should be fixed now!

dirkluijk:
I tried the signature
```typescript
get<R extends T, T = any>(type: Token<T>, fromComponentInjector = false): R & SpyObject<R> {
```
However, because of the conditional types I have issues with extending classes. See https://github.com/Microsoft/TypeScript/issues/26627.

I feel more comfortable with the old signature, but then you can use any token (not restricted to super type).

I will take a deeper look later this week.

NetanelBasal:
We need to support the following:

```
{
  provide: SomeSuperClass,
  useClass: SomeSubClass
}
```

Based on the above provider:

```ts
spectator.get(SomeSuperClass) // should return `SomeSuperClass`
spectator.get<SomeSubClass>(SomeSuperClass) // should return `SomeSubClass`
```

dirkluijk:
Thanks for deciding @NetanelBasal. 😄 I 'll try to make it working.

NetanelBasal:
It's already work in this branch. (at least with Jasmine)

projects/spectator/jest/src/host.ts
```diff
@@ -1,31 +1,9 @@
...
-   */
-  get<T>(type: Token<T>, fromComponentInjector = false): T & SpyObject<T> {
-    return super.get(type, fromComponentInjector) as T & SpyObject<T>;
-  }
-}
+import { createHostComponentFactory as baseCreateHostComponentFactory, HostComponent, SpectatorOptions, SpectatorWithHost as BaseSpectatorWithHost } from '@netbasal/spectator';
+import { CreateComponentOptions } from '../../src/lib/types';
 
-export function createHostComponentFactory<C, H = HostComponent>(options: SpectatorOptions<C, H> | Type<C>): (template: string, detectChanges?: boolean, complexInputs?: Partial<C>) => SpectatorWithHost<C, H> {
-  if (!isType(options)) {
```
dirkluijk: @NetanelBasal Why did you remove this block? It is needed to create Jest compatible spies (the whole point of extending in this secondary entry point) ;)

projects/spectator/jest/src/service.ts
```diff
@@ -1,28 +1,13 @@
...
 }
 
-export function createService<S>(options: ServiceParams<S> | Type<S>): SpectatorService<S> {
-  if (!isType(options)) {
-    options.providers = options.providers || [];
-    (options.mocks || []).forEach(type => options.providers.push(mockProvider(type)));
-    options.mocks = [];
-    options.declarations = options.declarations || [];
-    options.entryComponents = options.entryComponents || [];
-  }
```
dirkluijk: Same here

projects/spectator/jest/src/host.ts
NetanelBasal: Because I don't understand why we need it? the `baseCreateHostComponentFactory` should take care of it.

NetanelBasal: It's for the `  (options.mocks || []).forEach(type => options.providers.push(mockProvider(type)));`?

dirkluijk: Yes

The trick is that this one uses the `mockProvider()` definition of `jest/mock.ts` instead of the base one. :)

NetanelBasal: Sorry, it was late :) You can re-add it.

dirkluijk: Would be nicer if we could pass this overridden function in a cleaner way,

e.g.

```typescript
function someCreateSpyFunctionForExampleForJest() {}

createHostComponentFactory({
   createSpyObject: someCreateSpyFunctionForExampleForJest
})
```

What do you think?

NetanelBasal: LGTM

dirkluijk: All right, let me try something 😄

dirkluijk:
Can you review my latest two commits?

What I did:
* Added `mockProvider` option, in order to make the Jest entry point less weird
* Renamed `ServiceParams` to `ServiceOptions` to make it resemble `SpectatorOptions`
* Changed a lot of `T & SpyObject<T>` because `SpyObject<T>` is already `T & SomethingElse`
* Fixed some generics

dirkluijk:
Sorry, last 3 commit. 😄 Let's squash in the end.

projects/spectator/jest/src/host.ts
dirkluijk: Should be fixed now

src/app/todos-data.service.spec.ts
```diff
@@ -33,7 +33,7 @@ describe('HttpClient testing', () => {
 
   it('should work with external service', fakeAsync(() => {
     const { dataService, expectOne, get } = http();
-    get(UserService).getUser.andCallFake(() => {
+    get<SpyObject<UserService>>(UserService).getUser.andCallFake(() => {
```
NetanelBasal: Why do we need to add the `SpyObject` here?

dirkluijk: Http factory does not support passing `mocks` property. So by default no providers are mocked. That is the reason `SpectatorHTTP` returns just type `S` instead of `SpyOject<S>`.

In this test we explicitly add a `mockProvider` ourselves so we know what we are doing.

I was planning on adding automatic mocking for HTTP as well, I'll get back to you about that.

dirkluijk:
There is a bit of inconsistency in the README examples:
```ts
// spectator
const createComponent = createTestComponentFactory(ButtonComponent);
let spectator: Spectator<ButtonComponent>;

// with-host  
const createHost = createHostComponentFactory(ZippyComponent);
let host: SpectatorWithHost<ZippyComponent>;

// etc
```

Can we change all examples to:
```ts
// Spectator
const createComponent = createTestComponentFactory(ButtonComponent);
let spectator: Spectator<ButtonComponent>;

// With Host  
const createComponent = createHostComponentFactory(ZippyComponent);
let spectator: SpectatorWithHost<ZippyComponent>;
```
This way, it is consistent with the new factories for service, HTTP and routing:
```ts
// Service
const createService = createServiceFactory(MyService);
let spectator: SpectatorService<MyService>;

// HTTP
const createHttp = createHttpFactory(MyDataComponent);
let spectator: SpectatorHttp<MyDataComponent>;

// Routing
const createComponent = createRoutedComponentFactory(MyPageComponent);
let spectator: SpectatorWithRouting<MyPageComponent>;
```
What do you think?

NetanelBasal:
Yes. I think we need to talk in Gitter. It's hard to communicate here.

dirkluijk:
Updated list with Spectator v4 changes:

## Features
* New `createServiceFactory` with `SpectatorService` and support for all base options
* New `createHttpFactory` with `SpectatorHttp`  and support for all base options
* New `createRoutedComponentFactory` with `SpectatorWithRouting` and support for all base options
* Improved way of overriding options
* Angular 8 upgrade
* Jasmine 3 support

## BREAKING CHANGES
* changed signature of the factories of `Spectator` and `SpectatorWithHost`

Before:
```ts
const spectator = createComponent({ foo: 'bar' }, false);
```
After:
```ts
const spectator = createComponent({ props: { foo: 'bar' }, detectChanges: false });
```

* deprecated `createHTTPFactory` in favour of the new `createHttpFactory`

Before:
```ts
const http = createHTTPFactory(SomeService);

it('...', () => {
  const { dataService, get } = http();
  dataService.foo();
});
```ts
After:
```
const createHttp = createHttpFactory(SomeService);

it('...', () => {
  const spectator = createHttp();
  spectator.dataService.foo();
});
```

* deprecated `createService` in favour of the new `createServiceFactory`

Before:
```ts
const service = createService(SomeService);

it('...', () => {
  service.foo();
});
```ts
After:
```
const createService = createServiceFactory(SomeService);

it('...', () => {
  const spectator = createService();
  spectator.service.foo();
});
```

dirkluijk:
I successfully upgrades a very large Angular project to Spectator 4 RC1. Let's see how that did go.

Migrations:
* **import fixes**: easy. Just a simple global replacement. We can easily make a migration for that as well.
* **new factory signatures (overrides)**: that took quite some effort. Is hard to do with a global replacement. For example: `createComponent(true, { /* some complex inline object */ })` or `createComponent('...template...', { /* some complex inline object */ })`.
* new factory names: easy. Just a global replacement.

Bugs:
* **Spectator**: one test case did fail because the removal of `getDirectiveInstance` was not yet documented. Replacing this with `query(MyDirective)` revealed that querying for directives like that was not working at all. I fixed that in https://github.com/ngneat/spectator/pull/147/commits/e3ff1fac56bfa0ffd5ad25a352986cf57c86c1fd.
* **SpectatorDirective**: fixing the previous issue revealed that the directive instance of `SpectatorDirective` was incorrect and pointed to the host component instead. Fixed in https://github.com/ngneat/spectator/pull/147/commits/bee85ddbc8238a5e125c243bb312a2d9945403ad.
* **SpectatorHost**: some test cases did fail because `hostFixture` was accidentally removed. Fixed in https://github.com/ngneat/spectator/pull/147/commits/84b2910d45681993cacb1dd2ed94cf0f0f7d5a63
* **SpectatorHttp**: replacing the old `createHTTPFactory` with `createHTTPFactory` revealed that destructuring resulted in a broken `expectOne` function. As destructuring was common in v3 and I wanted to avoid issues, I added a small fix to support this in https://github.com/ngneat/spectator/pull/147/commits/51157902865afb8e3f3d2c94b1da048722aad452

Docs:
* Published `MIGRATION.md` to this branch (and added to build script)
* Added the removal of `getDirectiveInstance` in the MIGRATION.md

PR #148, opened 2019-08-07
fix: fix query regression and unit testing improvements

dirkluijk:
OK, so I was a bit stubborn and decided to try to fix the regression.

In this PR:
- The regression was caused by the `SpectatorDebugElementNotFoundError` error which was thrown and catched immediately - but not in all cases. I removed the throw-and-catch mechanism in favour of a silent return of `null` (which is expected in all cases, especially for matchers like `toExist()`.
- In order to see the coverage of `projects/spectator` instead of `src`, I needed to move the "test" builder to the `spectator` library project. I did not move any test files around and decided to postpone a complete new test setup for a later moment (soon). For the current coverage, see the screenshot below.
- Added some `@deprecated` annotations for untested files.

image

I think this fixes the issues in https://github.com/NetanelBasal/spectator/pull/147 as well. I'll take a look.

Closes https://github.com/NetanelBasal/spectator/issues/146 (added additional test case).

I don't see any issues anymore?


package.json
```diff
@@ -6,9 +6,8 @@
     "ng": "npx ng",
     "start": "npx ng serve",
     "build": "npx ng build",
-    "test": "npx ng test",
+    "test": "npx ng test spectator",
     "test:jest": "npx ng run spectator-app:test-jest",
-    "test:jest:watch": "npx ng run spectator-app:test-jest --watch",
```
NetanelBasal: Why you removed the watch option?

projects/spectator/src/lib/internals.ts
```diff
@@ -290,13 +281,7 @@ export class Spectator<C> {
...
 export function _getChild<R>(debugElementRoot: DebugElement) {
   return function(directiveOrSelector: Type<R> | DOMSelector | string, options: { read } = { read: undefined }): R {
-    const [child] = _getChildren<R>(debugElementRoot)(directiveOrSelector, options);
-
-    if (!child) {
-      throw new SpectatorDebugElementNotFoundError(`Cannot find a debug element for ${directiveOrSelector}`);
-    }
-
-    return child;
+    return _getChildren<R>(debugElementRoot)(directiveOrSelector, options)[0] || null;
```
NetanelBasal: I think we should at least add console info so that users will know what is missing.

projects/spectator/src/lib/globals.ts
```diff
@@ -1,7 +1,17 @@
+/**
+ * @deprecated Deprecated in favour of Spectator query methods and Spectator custom matchers
+ *
+ * To be removed in v4
+ */
```
NetanelBasal: 💯

projects/spectator/src/lib/mock-directive.ts
```diff
@@ -5,6 +5,7 @@ import { Directive, EventEmitter, Type } from '@angular/core';
  * MockDirective({ selector: 'some-directive' });
  * MockDirective({ selector: 'some-directive', inputs: ['some-input', 'some-other-input'] });
  *
+ * @deprecated Deprecated in favour of the `ng-mocks` implementation of `MockDirective`. To be be removed in the next major version.
```
NetanelBasal: Removed in v4 PR.

package.json
dirkluijk: You can run both ` test` and `test:jest` in watch mode with `--watch` so it felt a bit unneeded. ;-)

projects/spectator/src/lib/internals.ts
dirkluijk: But that was before also not the case, right? Because the error was catched?

projects/spectator/src/lib/mock-directive.ts
dirkluijk: 🚀

projects/spectator/src/lib/internals.ts
NetanelBasal: There was `throw error`

dirkluijk: Oh, sorry, I see. I'll fix it.

dirkluijk: Oh, well, it is still thrown right? Because it is not catched?

NetanelBasal:
@dirkluijk, please merge it to master and release a new version that fixes the regression.

dirkluijk:
Of course.

dirkluijk:
Released in 3.11.1.

PR #149, opened 2019-08-08
refactor: new lib structure with internals and new factories

dirkluijk:
So I did a little bit of refactoring.

I had this idea:
* `BaseSpectator` (internal)
* `Spectator extends BaseSpectator`
* `SpectatorWithHost extends Spectator`

So that we can add in the future:
* `ServiceSpectator extends BaseSpectator` (with factory, options, etc)
* `RoutedSpectator extends Spectator` (with factory, options, etc)
* `HttpSpectator extends BaseSpectator` (with factory, options, etc)

I did not implement it exactly like that (yet), but did some preparation.

In order to have a common base, I did an internal restructure of the lib. Besides that, I moved internal functions to  `internal` directory.

This PR does not yet introduce breaking changes as far as I know. No unit tests fail.


projects/spectator/src/lib/spectator-with-host/create-factory.ts
```diff
@@ -0,0 +1,76 @@
...
+  const moduleMetadata = initialModule<Component, Host>(typeOrOptions);
+
+  const factoryCD = isType(typeOrOptions) || typeOrOptions.detectChanges === undefined ? true : typeOrOptions.detectChanges;
+  const componentProviders = isType(typeOrOptions) ? [] : typeOrOptions.componentProviders || [];
+
+  beforeEach(() => {
+    jasmine.addMatchers(customMatchers as any);
+    TestBed.configureTestingModule(moduleMetadata);
+
+    if (componentProviders) {
```
NetanelBasal: This will always be true

NetanelBasal: You need .length

projects/spectator/src/lib/spectator/host-component.ts
```diff
@@ -0,0 +1,6 @@
+import { Component } from '@angular/core';
```
NetanelBasal: should be in spectator-with-host-folder

projects/spectator/src/lib/spectator-with-host/spectator-with-host.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  queryHost<R extends Element>(selector: string | DOMSelector, options?: { root: boolean }): R;
+  queryHost<R>(directive: Type<R>): R;
+  queryHost<R>(directiveOrSelector: Type<any> | string, options: { read: Token<R> }): R;
+  queryHost<R>(directiveOrSelector: QueryType, options?: QueryOptions<R>): R {
+    if ((options || {}).root && isString(directiveOrSelector)) {
+      return document.querySelector(directiveOrSelector) as any;
+    }
+    return _getChildren<R>(this.hostDebugElement)(directiveOrSelector, options)[0] || null;
```
NetanelBasal: We can move `_getChildren` to internal and `_setInput` to internal as well.

dirkluijk: 👍

projects/spectator/src/lib/spectator/host-component.ts
dirkluijk: Should we? Because it is the default Host for Spectator.

projects/spectator/src/lib/spectator-with-host/create-factory.ts
dirkluijk: Check!

projects/spectator/src/lib/spectator/host-component.ts
NetanelBasal: `createTestComponentFactory` doesn't use it.

dirkluijk:
@NetanelBasal I spend a few hours trying to improve some things and test some new things. I am very interested in your opinion! 😄 

## What has been changed?

### Internal refactoring

* Defined shared interfaces, `BaseSpectator`, `BaseOptions`, etc (internal)
* More consistent file names
* Improved options handling
* TyoeScript `strictNullChecks` for library
* Marked some functions `@internal` or `@publicApi`
* TSLint rules (e.g. ordered-imports)
* Removed the use of `EntryComponentModule` (looks like declaring it in the test module works as well)

### API changes

Just a few.

* Made `initialModule` and related functions internal
* Changed `SpectatorService()` constructor to `SpectatorService(service)` (can be reverted if needed)
* Removed `SpectatorWithHost.getDirectiveInstanceSomeDirective()`  in favour of `SpectatorWithHost.queryHost(SomeDirective)` (can be changed into @deprecated if needed)
* Renamed `CreateComponentOptions` to `SpectatorOverrides` and renamed its property `props` to `properties` (was new in v4)

### Consistent factories

For `Spectator` and `SpectatorWithHost` nothing has been changed on the outside.

#### createServiceFactory()

The `createService()` function has been deprecated in favour of `createServiceFactory()`
which returns a factory for `SpectatorService`:  

```typescript
const createService = createServiceFactory(MyService);

it ('should work', () => {
  const spectator = createService();
  
  spectator.service.foo();
  spectator.get(OtherService).bar();
})
``` 

Or:

```typescript
const createService = createServiceFactory({
  service: MyService,
  mocks: [OtherService],
  providers: []
});

it ('should work', () => {
  const spectator = createService({
    providers: [] // override providers, like `Spectator` and `SpectatorWithHost`
  });
  
  spectator.service.foo();
  spectator.get(OtherService).bar();
})
``` 
#### createHttpFactory()

// todo

### `SpectatorFactory`

I don't know if this adds any value. But since we have several factories which share the same API characteristics, I added a `SpectatorFactory` object:

```typescript
SpectatorFactory.forComponent(FooComponent);
SpectatorFactory.forComponentWithHost(BarComponent);
SpectatorFactory.forService(LoremService);
SpectatorFactory.forHTTP(MyDataService);
```
##### Example
```
const createComponent = SpectatorFactory.forComponent(FooComponent);
// same as: const createComponent = createTestComponentFactory(FooComponent);

it('should create component', () => {
    const spectator = createComponent();
});

const createHost = SpectatorFactory.forComponentWithHost(BarComponent);
// same as: const createHost = createHostComponentFactory(BarComponent);

it('should create component with host', () => {
    const spectatorWithHost = createHost();
});

const createService = SpectatorFactory.forService(LoremService);
// same as: const createService = createServiceFactory(LoremService);

it('should create service', () => {
    const spectator: SpectatorService<LoremService> = createService();
});

// not yet implemented
const createHTTP = SpectatorFactory.forHTTP(MyDataService);
// same as: const createHTTP = createHTTPFactory(MyDataService);

it('should create data service', () => {
    const spectator: SpectatorHttp<MyDataService> = createHTTP();
});
```

dirkluijk:
Fixed tests. I'll wait for your opinions on this matter before merging to v4 branch.

NetanelBasal:
LGTM. Few notes:

- Change `getDirectiveInstance` to `queryHost(SomeDirective)`. (No need for deprecation)
- Keep the name `props` instead of `properties`. It's shorter and common in the community.
- I like the consistent API. Deprecate `createService`, and HTTP in favor of the new factories. (Or maybe keep both?)
- I can't see any reason to keep the `SpectatorFactory` helpers.
- Keep in mind that we need to update the CLI and schematics.

NetanelBasal:
@dirkluijk TestingModule doesn't support `entryComponents` out of the box. That's why I added this "hack".

dirkluijk:
Thanks for your feedback, I will push some changes later today.

> I like the consistent API. Deprecate createService, and HTTP in favor of the new factories. (Or maybe keep both?)
I will keep them both. I think there is no direct need to deprecate createService (it just works differently), but I will deprecate HTTP because it does look a lot like the new factory and may confuse people.

> TestingModule doesn't support entryComponents out of the box. That's why I added this "hack".
But the test seems to pass. Can you investigate it? ;)

NetanelBasal:
Sure, I will check it when we finish with the rest. I have another proposal. I'm thinking about removing the `SpectatorWithHost` entity and merge it somehow with `createTestComponentFactory`. So, maybe something like:

```ts
let spectator: Spectator<ButtonComponent>;

const createComponent = createTestComponentFactory({
  component: ButtonComponent,
  template: ``,
  host: Host
});
```

We can use typescript to infer the return type based on the `template` property.  
There will probably more edge cases we need to check, but I think it will make the API much more straightforward and cleaner. What do you think?

NetanelBasal:
We can even call `createHostComponentFactory` under the hood when someone passes a template.

dirkluijk:
Yes, I was already thinking of something like that! They share a lot...
shall I do an attempt for it?

NetanelBasal:
💯

dirkluijk:
* Reverted ` getDirectiveInstance` (but changed its internals to use query)
* Reverted rename of props
* Removed SpectatorFactory helpers.

NetanelBasal:
Reverted getDirectiveInstance? why?

> Change getDirectiveInstance to queryHost(SomeDirective). (No need for deprecation)

dirkluijk:
Oh, sorry, I misread your comment. ;)

dirkluijk:
About:

```
export interface SpectatorOverrides<Component> extends BaseSpectatorOverrides {
  detectChanges?: boolean;
  props?: Partial<Component> & HashMap;
}

interface HashMap<T = any> {
  [key: string]: T;
}
``` 

This means that `props` is `Partial<Component> & { [key: string]: any }` which is (almost!) same as `{ [key: any]: any }` (except you can't add methods that doesn't exist).

Why not just `Partial<Component>`? This prevents adding custom properties, which (I think) is a good thing.

NetanelBasal:
There were cases when I needed to add more properties to the host component.

NetanelBasal:
I will show you an example later on.

NetanelBasal:
> OK. I think you loose valuable type-safety there. 😉 But I'll keep it.

So, sometimes it's overkill to create a custom host. A quick workaround could be:
```ts
host = createHost(`<div [someDirective]="{ value: dynamic }"></div>`, true, {
  dynamic: 'value'
});
```

Maybe we could add one more option - `hostProps`, which will be `HashMap<any>`.

dirkluijk:
Yeah, but the options you are passing are applied on the Component (C), not on the HostComponent (H). And C is always known.

So yeah, I would then go for a restricted props and a non-restricted hostProps. 👍 
I will add it to my todo list 😄 

I did more refactoring. I discovered that `Spectator` did use generic types for HostComponent, but that didn't make sense. `SpectatorWithHost` is now a pure extension of `Spectator`.

I will now do that attempt to make the `createTestComponentFactory` more generic.

NetanelBasal:
Great. Thanks for all your excellent work. I'm currently busy working on the next open-source library :)

NetanelBasal:
The plan with v4 is also moving this repo to a new organization and not my personal scope `@netbasal`.

NetanelBasal:
Version 4 is going to be awesome. We'll publish a blog post about what's new. I expect more developers to adopt this library.

dirkluijk:
Yeah, Spectator is an awesome library with a lot of type-safety. I really love it.
v4 will be AWESOME. Wait until you see my SpectatorWithRouting PR soon 😎 

I made this small library last week: https://github.com/dirkluijk/observable-matchers
And I am planning on something for creating type-safe test data as wel.

I added the host props + unit tests, see my latest commit.

NetanelBasal:
observable-matchers looks good. We can mention it in the new docs.

> I added the host props + unit tests, see my latest commit.

LGTM, thanks.

dirkluijk:
I have been thinking about your idea:

> Sure, I will check it when we finish with the rest. I have another proposal. I'm thinking about removing deprecating the SpectatorWithHost entity and merge it somehow with createTestComponentFactory. 

I understand your idea, and tried some things, but I too much issues:

* Clarity: testing a component standalone vs. with a host serve different purposes and it is powerful to have a different API for overrides and options
* Code quality: I have successfully managed to keep the inheritance very pure and I must say, I like the current separation
* API changes: the current design has no breaking changes and resembles the current API. If we move away from this, we might make things a bit confusing for users. Agreed, it does not need to be breaking, but what factory are users *supposed* to use? Keeping the existing host factory non-deprecated might make things even more confusing.
* Extension by other factories: I am currently preparing a routed component factory (`SpectatorWithRouting`) which extends the `Specator`  but does not share the charactistics of `SpectatorWithHost`, because a routed component should not have a host interaction.

So, my proposal for v4:

* Merge this PR to the `v4` branch
* I would like to spend some time on the unit tests suite of Spectator (maybe even rewrite it)
* I would like to have my `SpectatorWithRouting` PR finished this weekend - including tests
* docs + schematics (if no one volunteers I will keep this for the end)

What do you think?

NetanelBasal:
Sure, Agreed. We can keep it like today. 

I think we should delete most of the specs. There is a mess there. We should create one component per feature that covers all the things we support. 

`WithoutHost`, `withHost`, `withCustomHost`, Service, HttpService.

I will do the schematics. I want to integrate it with the default angular commands instead of how it works now.

dirkluijk:
Awesome! Yeah I can help with the docs next week.

I will merge this PR into v4 and do a first test against an existing before continuing with unit tests + routing.

PR #157, opened 2019-08-19
feat: add factory for SpectatorForDirective

dirkluijk:
So what I basically did:

- Moved stuff from `Spectator` to abstract `DomSpectator` (for querying, event dispatching, etc) in order to reuse it
- Created new `SpectatorForDirective` which contains stuff from both `Spectator` and `SpectatorWithHost`
- it uses the host fixture, so that querying etc is done from host
- instead of `component` this one has a `directive` instance
- existing code still works (I kept the unit tests)
- added some tests for the new factory (we should have more of them I think)
- updated README

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,9 @@ describe('With Custom Host Component', function () {
 });
 ```
 ## Testing Directives
+
+There is a special test factory for testing directives.  
+
 Let's say we have the following directive:
 
 ```ts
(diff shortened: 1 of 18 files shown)


projects/spectator/jest/test/highlight.directive.spec.ts
```diff
@@ -24,3 +24,25 @@ describe('HighlightDirective', () => {
...
   });
 });
+
+describe('HighlightDirective (createHostDirectiveFactory)', () => {
+  let host: SpectatorForDirective<HighlightDirective>;
+
+  const createHost = createHostDirectiveFactory(HighlightDirective);
+
+  // calculated styles not supported in JSDOM
+  xit('should change the background color', () => {
```
NetanelBasal: remove xit

README.md
```diff
@@ -278,6 +278,9 @@ describe('With Custom Host Component', function () {
 });
 ```
 ## Testing Directives
+
+As of Spectator v4, there is a special test factory for testing directives.
```
NetanelBasal: This readme will only be visible from v4, so there is no need to mention it.

projects/spectator/jest/test/highlight.directive.spec.ts
dirkluijk: Oh, I removed it but the tests fail. I had to exclude these tests earlier because the `toHaveStyle()` matcher is not supported in Jest. It relies on `getComputedStyle().getPropertyValue()`.

NetanelBasal: There is a way to check if the user uses Jest and warn them that it will not work?

dirkluijk: We currently don't have a way to detect it. Let's see if another solution is possible, e.g. not relying on computed styles. Let's fix it in a separate PR.

PR #169, opened 2019-08-28
fix(vscode-snippets): change snippet language

Plysepter:
This enables the use of the code snippets within Typescript files.


dirkluijk:
Awesome, thanks!

PR #178, opened 2019-09-04
feat(overrides): improve type inference for host props overrides

dirkluijk:
## PR Checklist

## PR Type

```
[X] Feature
```

## What is the new behavior?
This PR improves type inference, so that the following is now possible (and type-safe):

```ts
class DummyComponent { public dummy = 'dummy'; }

const createHost = createHostFactory({
    component: DummyComponent
});

const spectator = createHost({
    hostProps: {
        foo: 'foo'
    }
});

spectator.host.foo = 'x'; // compiles
spectator.host.bar = 'x'; // does NOT compile
spectator.component.dummy = 'x'; // compiles
```

When using a custom host component (e.g. a component that does not extend `HostComponent`), the passed host props should still extend the custom host class.

```
[ ] Yes
[X] No
```


dirkluijk:
Additionally, this PR fixes the following.

* Previously, when the debugElement of `SpectatorHost` or `SpectatorDirective` could not be found, it used the host's debugElement as fallback (including its `componentInstance`)
* This can lead to weird behaviour, like `spectator.component` to be the `HostComponent`, or applying the `props` to the HostComponent, which is certainly wrong.
* When removing this fallback, there was one unit testing failing, regarding structural directives. Why, you might ask? Well, structural directives don't have a `DebugElement` as they are not bound to any element but rather an ng-template. Therefore, they are only bound to a `DebugNode`.
* To fix this, this PR tries to find the instance by looking for a `DebugNode`.
* If the instance can still not be found, an error is now thrown. Of course, this might be considered a breaking change. But I would still vote for having it, as `SpectatorHost`/`SpectatorDirective` did never support missing debug elements / instances but silently allowed it but with weird side effects.

dirkluijk:
@NetanelBasal This PR is now as I would like to have it. Pure, clean type-inference without typecasting or weird hacks. 

However, type inference is hard to test in a unit test, as expected compile errors cannot be tested with our current tooling.

I tested it manually by uncommenting the lines in the newly added unit tests. Can you double check against your project(s)?

projects/spectator/src/lib/spectator-host/create-factory.ts
```diff
@@ -92,15 +89,29 @@ export function createHostFactory<C, H = HostComponent>(typeOrOptions: Type<C> |
...
+  props?: Partial<C>,
+  hostProps?: HP
+): SpectatorHost<C, H & HP> {
   const hostFixture = TestBed.createComponent(options.host);
-  const debugElement = hostFixture.debugElement.query(By.directive(options.component));
+  const debugElement = hostFixture.debugElement.query(By.directive(options.component)) || hostFixture.debugElement;
+  const debugNode = hostFixture.debugElement.queryAllNodes(nodeByDirective(options.component))[0];
+
+  if (!debugNode) {
+    throw new Error('Cannot find component/directive in host template');
```
NetanelBasal: Maybe add some emojis to the error? 😀

projects/spectator/src/lib/spectator-host/spectator-host.ts
```diff
@@ -57,7 +54,7 @@ export class SpectatorHost<C, H = HostComponent> extends Spectator<C> {
 
   public setHostInput<K extends keyof H>(input: Partial<H>): void;
   public setHostInput<K extends keyof H>(input: K, inputValue: H[K]): void;
-  public setHostInput<K extends keyof H>(input: Partial<H> | K, value?: H[K]): void {
+  public setHostInput(input: any, value?: any): void {
```
NetanelBasal: There is no type safety in `setHostInput` anymore?

projects/spectator/jest/test/override-typesafety.component.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+      host = createHost('<my-component></my-component>', {
+        hostProps: {
+          // control: new FormControl(), // should not compile
+          foo: 'x'
+        }
+      });
+    });
+
+    it('should allow setting the defined properties', () => {
+      host.hostComponent.foo = 'bar';
```
NetanelBasal: We can try this tool: https://github.com/microsoft/dtslint. I know that RxJS using it.

projects/spectator/src/lib/spectator-host/create-factory.ts
NetanelBasal: And the name of the component/directive

NetanelBasal:
> I tested it manually by uncommenting the lines in the newly added unit tests. Can you double check against your project(s)?

I will.

projects/spectator/jest/test/override-typesafety.component.spec.ts
dirkluijk: That is definitely something for on the roadmap! 😄 Thanks for the idea.

projects/spectator/src/lib/spectator-host/spectator-host.ts
dirkluijk: When using overloads, the actual function implementation signature does not need to be typed,  at least not for the outside world. As long as it is compatible with the overloads. The typings in the implementation signature are only used for inside the implementation.

projects/spectator/src/lib/spectator-host/create-factory.ts
dirkluijk: Done

projects/spectator/src/lib/spectator-host/spectator-host.ts
NetanelBasal: Yes, I know. Sorry, missed the overloads :)

dirkluijk: I made the signature simplier because it would otherwise be harder to match the improved signature of setProps 😄

PR #190, opened 2019-09-28
fix: initial change detection before querying debug element

dirkluijk:
This PR allows to have tests with `SpectatorHost` and `SpectatorDirective` where an inital change detection needs to run, before the component/directive is actually visible.

Example:
```ts
createComponent(`
    <mat-horizontal-stepper>
        <mat-step>
            <app-stepper-controls></app-stepper-controls>
        </mat-step>
    </mat-horizontal-stepper>
`);
```
(where `app-stepper-controls` is the component under test)

This behaviour can be overriden using the existing `detectChanges` option (not sure if that makes sense).

### DIFF
--- a/projects/spectator/src/lib/spectator-directive/create-factory.ts
+++ b/projects/spectator/src/lib/spectator-directive/create-factory.ts
@@ -66,7 +66,7 @@ export function createDirectiveFactory<D, H = HostComponent>(
       set: { template }
     });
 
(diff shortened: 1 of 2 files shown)


dirkluijk:
I see one test is still failing. Will take a look at it tomorrow.

dirkluijk:
Closed. I had a different solution in mind, but somehow I cannot reproduce my original issue anymore. 😄

Issue #202, opened 2019-10-10
Add selectOption helper


NetanelBasal:
Create a helper that select a value in a select element (support multi):

`spectator.selectOption(selectSelector, id/s)`

Coly010:
I'll give this a whirl.

PR #208, opened 2019-10-13
feat(spectator): add selectOption helper method

Coly010:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

There is not currently an API to select options within `<select></select>`.

Issue Number: #202 

## What is the new behavior?

Add `selectOption` helper to select options in a `<select></select>`
Add `toHaveSelectedOptions()` matcher.

Usage:
```ts
it('should set the correct options on multi select', () => {
  const select = spectator.query('#test-multi-select') as HTMLSelectElement;
  spectator.selectOption(select, ['1', '2']);
  expect(select).toHaveSelectedOptions(['1', '2']);
});

it('should set the correct option on standard select', () => {
  const select = spectator.query('#test-single-select') as HTMLSelectElement;
  spectator.selectOption(select, '1');
  expect(select).toHaveSelectedOptions('1');
});
```

```
[ ] Yes
[x] No
```


benelliott:
This looks great - would there be any room to make the feature less assuming about what specific select-like implementation you are using? 

I am just anticipating that a lot of users will be trying to apply this to a [`MatSelect`](https://material.angular.io/components/select/overview).

This might be out of scope for this PR as finding a `MatSelect`'s options involves some fiddling with dialog containers, but worth keeping in mind.

Coly010:
I would think supporting a specific UI library would be out of scope, however, it could be possible to add an optional parameter to the `selectOption` method for the `options` selector.

However, if the process for selecting the `mat-option` involves simulating a click on the `mat-select` before it can access the `mat-option` then to me that would be a highly specific process.

I feel if that support was to be offered with `spectator` then it could be a as a plugin such as `spectator-angular-material-support` which could contain a lot of helper methods and matchers specifically for the Angular Material library.

benelliott:
@Coly010 Definitely agree that Material-specific stuff shouldn't be in this package. I just meant it would be great if the feature allowed you to override the usage of native `select` where possible.

Coly010:
I feel like allowing the override of the native select would require the dev writing the test to write some implementation specific code to handle their specific select-like solution, which would make the usage of this helper method redundant, surely, as the logic to select the options required would still end up being written by the dev in their test suites?

benelliott:
Yeah I think you're right. Adding a plugin for a specific implementation looks like the only way to do this.

NetanelBasal:
@benelliott I also thought of the same thing but in a more general way. I mean support for custom select elements. But then I realized that you should know what you need to open, what you need to click on, and how to assert expected value. 

Sure, we can make some helpers, but in this case, the developers can benefit from our API. (unless you have some cool idea)

@Coly010 awesome work!

projects/spectator/src/lib/matchers-types.d.ts
```diff
@@ -41,5 +41,7 @@ declare namespace jasmine {
     toHaveDescendant(selector: string | Element): boolean;
 
     toHaveDescendantWithText({ selector, text }: { selector: string; text: string }): boolean;
+
+    toHaveSelectedOptions(expected): boolean;
```
NetanelBasal: `expected: string | string[]`

projects/spectator/src/lib/select-option.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    setOptionSelected(option, element);
+  } else {
+    if (!element.multiple) {
+      return;
+    }
+
+    element.querySelectorAll('option').forEach(opt => {
+      if (options.includes(opt.value)) {
+        setOptionSelected(opt, element);
```
NetanelBasal: Do we need to dispatch the `change` event for each selected option?

Coly010: I figured that a developer might have an `onChange` handler set up that they are trying to test when certain options are selected.

NetanelBasal: What's the behavior when using the native JS API?

Coly010: There is no native way of selecting multiple values of a select at once. From looking around online, a lot of solutions to this involve setting the selected option to true. 

It doesn't fire the change event simply by doing this. However, by firing the change event, I feel it's closer to simulating how a user would select the multiple values.

Perhaps this should be an option for the developer that they can turn off when using the method.

NetanelBasal: > Perhaps this should be an option for the developer that they can turn off when using the method.

Agree.

projects/spectator/src/lib/matchers-types.d.ts
Coly010: Fixed.

projects/spectator/src/lib/select-option.ts
Coly010: Added optional parameter and two more tests to cover it's usage.

projects/spectator/test/form-select/form-select.component.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    expect(select).toHaveSelectedOptions(['1', '2']);
+    expect(onChangeSpy).toHaveBeenCalledTimes(2);
+  });
+
+  it('should not dispatch correct number of change events', () => {
+    const onChangeSpy = spyOn(spectator.component, 'handleChange');
+    const select = spectator.query('#test-onchange-select') as HTMLSelectElement;
+
+    spectator.selectOption(select, ['1', '2'], false);
```
NetanelBasal: Oh sorry, I wrote a comment a week ago, and just noted I didn't click the submit button.

Can we change it to `spectator.selectOption(select, ['1', '2'], { emitEvent: true });` so that it'll be scalable and readable, please?

Coly010: Sure thing!

Coly010: Changed

NetanelBasal:
This shouldn't be committed as a breaking change:

```
### BREAKING CHANGES

* **spectator:** Change from a boolean argument to an object argument
```

It causes the standard version to bump to v5. I'll see how to fix it.

Coly010:
> This shouldn't be committed as a breaking change:
> 
> ```
> ### BREAKING CHANGES
> 
> * **spectator:** Change from a boolean argument to an object argument
> ```
> 
> It causes the standard version to bump to v5. I'll see how to fix it.

My apologies. I had assumed breaking change as it was a change in the `selectOption` API. 
Would amending the HEAD commit at master and force-pushing fix this?

NetanelBasal:
I'm not sure. That's OK. Later, I will see how to fix it.

NetanelBasal:
@Coly010 I reverted the merge. Please, change the commit message on your branch and create the PR again.

PR #233, opened 2019-11-20
docs: add info on SCAM and declareComponent usage

MarcScheib:
## PR Checklist

- [ x ] The commit message follows our guidelines: https://github.com/ngneat/spectator/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ x ] Documentation content changes
[ ] Other... Please describe:
```

Add documentation on how to test components which are included in a module specific to that component. Its not necessary then to define the imports list again and instead use the module as it is.

It took me some time check the code if it is possible at all to avoide the described error and, thus, I found the property is already there but not described.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ When used like this, however, Spectator internally adds the component `ButtonCom
 Type ButtonComponent is part of the declarations of 2 modules [...]
 ```
 
-It is possible to tell Spectator not to add the component to the declarations of the internal module and, instead, use the manually defined module as is. Simply set the `declareComponent` property of the factory options to `false`:
+It is possible to tell Spectator not to add the component to the declarations of the internal module and, instead, use the explicitly defined module as is. Simply set the `declareComponent` property of the factory options to `false`:
 
 ```ts
 const createComponent = createComponentFactory({
--- a/README.md
+++ b/README.md
@@ -173,6 +173,34 @@ it('should work with tick', fakeAsync(() => {
 }))
 ```
 
+### Testing Single Component Angular Modules
+
+Components that are declared in their own module can be tested by defining the component 
+module in the imports list of the component factory together with the component. For example:
+
+```ts
+const createComponent = createComponentFactory({
+  component: ButtonComponent,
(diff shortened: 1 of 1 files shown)


benelliott:
OOC, what's the motivation for testing the module like this and not just the component?

MarcScheib:
I have component which rely on shared ones, i.e. my module is importing those for the component, Instead of specifying the imports in the spec again, I simply import the module with all its import and the declaration of the component.

E.g. if my component requires something like:
```
  imports: [
    CommonModule,
    MatButtonModule,
    MatIconModule,
    MatCardModule,
  ]
```

The spec would be like:
```ts
  const createComponent = createComponentFactory({
    component: ProfileComponent,
    imports: [CommonModule, MatButtonModule, MatIconModule, MatCardModule],
  });
```

As described in my PR, it could be (assume you have a module for the component):
```ts
  const createComponent = createComponentFactory({
    component: ProfileComponent,
    imports: [ProfileComponentModule],
    declareComponent: false,
  });
```

README.md
```diff
@@ -173,6 +173,34 @@ it('should work with tick', fakeAsync(() => {
...
+});
+```
+
+When used like this, however, Spectator internally adds the component `ButtonComponent` to the declarations of the internally created new module. Hence, you will see the following error:
+
+```
+Type ButtonComponent is part of the declarations of 2 modules [...]
+```
+
+It is possible to tell Spectator not to add the component to the declarations of the internal module and, instead, use the manually defined module as is. Simply set the `declareComponent` property of the factory options to `false`:
```
theblushingcrow: It is possible to tell Spectator not to add the component to the declarations of the internal module, and instead, use the explicitly defined module as is.


explicitly seems to me more accurate than manually.

NetanelBasal:
Please move it under the `Mocking Components` section, and add it to `Table of Contents`.

LayZeeDK:
I approve 👌

Issue #231, opened 2019-11-19
SpectatorHttp.expectOne fails if there are concurrent HTTP requests


Sturgelose:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

Trying to test multiple HTTP requests with `SpectatorHttp` and `expectOne` fails with 
`Error: Expected no open requests, found 1: GET request2`.
This seems to be some extra validation done in the `expectOne` function, when it shouldn't validate if there are other requests available to be flushed.

Seems to be specifically this line: https://github.com/ngneat/spectator/blob/master/projects/spectator/src/lib/spectator-http/spectator-http.ts#L47

## Expected behavior

I would expect a similar behaviour as Angular HttpMock, and allow to verify multiple Http requests done simultaneously. Either this can be given through a flag in the `expectOne` function or to just remove it and make the developer set it in the test itself.

Also, I haven't tested, but seems that this would also happen for any closed httpRequest unless you use `ignoreCancelled` parameter in `verify()`

## Minimal reproduction of the problem with instructions
Will try to provide a stackblitz, when I have time.

```typescript
// api-connector.service.ts
  public addRequests(requests: APIRequest[]) {

    // Setup a pipe for each request
    requests.forEach(request => {
      // Create dictionary entry to store the subject and the metadata
      this._requestSubject$[request] = { id: request, observable$: new Subject()}

      this._http
        .get(request)
        .subscribe(this._requestSubject$[request].result);
    });

  }

  public getResult(queryId: string): Observable<TimeSeries | DataPoint> {
    return this._requestSubject$[queryId].observable$;
  }
```

```typescript
// api-connector.spec.ts
it('should be created', () => {
    const request1 = 'request1';
    const request2 = 'request2';
    spectator.service.addRequests([request1, request2]);
    spectator.service.getResult(request1).subscribe((a) => console.log(a));
    spectator.service.getResult(request2).subscribe((a) => console.log(a));
    spectator.expectOne('request1', HttpMethod.GET).flush({hello: 'world'}); // FAILS
    spectator.expectOne('request2', HttpMethod.GET).flush({foo: 'bar'});

  });
```

## What is the motivation / use case for changing the behavior?

[...]

PR #238, opened 2019-11-28
 feat(spectator): added expectConcurrent to spectator-http (#231)

dedwardstech:
## PR Checklist

## PR Type
An addition to the SpectatorHttp API.

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`spectator.expectOne()` calls `this.controller.verify()` which causes tests to fail if there are concurrent requests going out.

Issue Number: 231

## What is the new behavior?
Adding `spectator.expectConcurrent()` allows for expecting multiple requests.

```
[ ] Yes
[x] No
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -739,7 +739,8 @@ describe('AuthService', () => {
 ```
 
 ## Testing with HTTP 
-Spectator makes testing data services, which use the Angular HTTP module, a lot easier. For example, let's say that you have service with two methods, one performs a GET and one a POST:
+Spectator makes testing data services, which use the Angular HTTP module, a lot easier. For example, let's say that you have service with three methods, one performs a GET, one a POST and one performs
+concurrent requests:
 
 ```ts
 export class TodosDataService {
@@ -751,6 +752,13 @@ export class TodosDataService {
 
   postTodo(id: number) {
     return this.httpClient.post('api/todos', { id });
+  }
+
+  collectTodos() {
+    return merge(
+      this.http.get('/api1/todos'),
+      this.http.get('/api2/todos')
+    );
   }
 }
 ```
@@ -777,6 +785,16 @@ describe('HttpClient testing', () => {
     const req = spectator.expectOne('api/todos', HttpMethod.POST);
     expect(req.request.body['id']).toEqual(1);
   });
+
+  it('can test current http requests', () => {
+    spectator.service.getTodos().subscribe();
+    const reqs = spectator.expectConcurrent([
+        { url: '/api1/todos', method: HttpMethod.GET },
+        { URL: '/api2/todos', method: HttpMethod.GET }
+    ]);
+    
+    spectator.flushAll(reqs, [{}, {}, {}]);
+  });
 });
 ```
 We need to create an HTTP factory by using the `createHttpFactory()` function, passing the service that you want to test. The `createHttpFactory()` returns a function which can be called to get an instance of SpectatorHttp with the following properties:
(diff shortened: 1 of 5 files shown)


dedwardstech:
A little rational behind adding this function.

Adding a flag to `expectOne` felt bulky and after a string of `expectOne(url, method, flag)`, you would then have to manually call `spectator.controller.verify()`. By adding this function you declare your expectations and the function will call verify for you, just like how `expectOne` works now.

README.md
```diff
@@ -786,6 +786,49 @@ We need to create an HTTP factory by using the `createHttpFactory()` function, p
...
+  beforeEach(() => spectator = createHttp());
+
+  it('can test current http calls', () => {
+    spectator.service.getTodos().subscribe();
+    const reqs = spectator.expectConcurrent([
+        { url: '/api1/todos', method: HttpMethod.GET },
+        { URL: '/api2/todos', method: HttpMethod.GET }
+    ]);
+
+    reqs.forEach(req => {
```
NetanelBasal: I like this approach, but I think we can make it cleaner. Something like: `flushAll()`

projects/spectator/test/todos-data.service.spec.ts
```diff
@@ -35,6 +35,21 @@ describe('HttpClient testing', () => {
     spectatorHttp.expectOne('two', HttpMethod.GET);
   });
 
+  it('should test concurrent requests', () => {
+    const spectatorHttp = http();
+
+    spectatorHttp.service.concurrentRequests().subscribe();
+    const reqs = spectatorHttp.expectConcurrent([
```
NetanelBasal: It can be cleaner: 
`spectatorHttp.expectConcurrent(['one', 'two', 'three'].map(url => 
    ({ url, method:HttpMethod.POST})))`

README.md
```diff
@@ -786,6 +786,49 @@ We need to create an HTTP factory by using the `createHttpFactory()` function, p
 - `get()` - A proxy for Angular `TestBed.get()`
 - `expectOne()` - Expect that a single request was made which matches the given URL and it's method, and return its mock request
 
+Spectator also makes testing concurrent requests easier. Similar to `spectator.expectOne()`, you can use,
```
NetanelBasal: Thanks for the effort, but I prefer to combine this example with the previous one. Please add the relevant code to previous snippet, and leave the method and the description here. Thanks.

README.md
dedwardstech: Are you talking about exposing another method on SpectatorHttp?

Possibly something like `spectator.flushAll(requests, [ { foo: 'lorem' }, { bar: 'ipsum' }])`?

NetanelBasal: Yes

dedwardstech: Okay, I'll make those changes.

dedwardstech:
I've added the flush all method, cleaned up the test for it and modified the documentation. For the documentation, I just merged the two together because it seemed to flow better.

I've given some thought to storing the pending requests for `expect(One/Concurrent)` in SpectatorHttp. It seems like it would make everything flow pretty nicely. Given that spectator is something of a convenience wrapper around the test bed, it would make sense to be able to call `spectator.flush()` or `spectator.flushAll()` instead of storing the requests in a variable and then flushing.

State could be stored pretty painlessly in spectator, but doing so might cause a breaking change. This is just something I thought about while working on this change. I'd be happy to put in a PR if this makes sense and there aren't things I'm missing here.

PR #266, opened 2020-02-11
feat: support for overriding mocking component view providers

AlesDo:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

It is not possible to override or mock the component's view providers using the create factory functions.

## What is the new behavior?

The view providers can be overridden using `componentViewProviders` and mocked using `componentViewMocks`.

```
[ ] Yes
[x] No
```


projects/spectator/src/lib/spectator-host/create-factory.ts
```diff
@@ -52,10 +52,16 @@ export function createHostFactory<C, H = HostComponent>(typeOrOptions: Type<C> |
     jasmine.addMatchers(customMatchers as any);
     TestBed.configureTestingModule(moduleMetadata);
 
-    if (options.componentProviders.length || options.componentMocks.length) {
+    if (
```
NetanelBasal: Could we extract it to a function and reuse it in both places, please?

AlesDo: I have extracted to a function and also added the possibility to just override only providers or viewProviders. After more thought I have seen that my original code would always override both even if you specified overrides only for one of them.
I also merged the tests together into the component provider tests since both are connected.

README.md
```diff
@@ -99,9 +99,11 @@ const createComponent = createComponentFactory({
   providers: [],
   declarations: [],
   entryComponents: [],
-  componentProviders: [], // Override the component's providers 
+  componentProviders: [], // Override the component's providers
```
NetanelBasal: We need to be consistent. Let's change it to `componentViewProvidersMocks`, please.

Issue #262, opened 2020-02-07
Support for Angular 9


fr0:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

```
warning " > @ngneat/spectator@4.11.1" has incorrect peer dependency "@angular/common@^8.0.0".
warning " > @ngneat/spectator@4.11.1" has incorrect peer dependency "@angular/core@^8.0.0".
warning " > @ngneat/spectator@4.11.1" has incorrect peer dependency "@angular/router@^8.0.0".
```

## Expected behavior

No peer dependency warnings when used with Angular 9.

For what it's worth, I've been using spectator with the beta/rc versions of Angular 9 for a while now and haven't encountered any issues other than the warnings.

NetanelBasal:
Sure, can you test it with the stable version and create a PR, please?

LayZeeDK:
We should also switch to TestBed.inject instead of TestBed.get.

codeNoobie:
@NetanelBasal Created https://github.com/ngneat/spectator/pull/267

Please review. I was getting a test failure with `should enable spying on properties` test inside the spyObject test suite. This was apparently happening as ts compilation for es6 was not allowing the getter property to be changed via spyOnProperty. I have fixed it by setting the test tsconfig target to es5. Please suggest if you know a better way.

NetanelBasal:
@codeNoobie thanks. Can you add `spectator.inject` method that'll call `TestBed.inject`, please?

codeNoobie:
Sure. Let me do that.

LayZeeDK:
Yes sorry, `TesbBed.get` is deprecated, not removed yet.

codeNoobie:
Not sure if Angular Team has officially marked it deprecated. But if it, then we should mark Spectator.get as deprecated too.

LayZeeDK:
https://angular.io/api/core/testing/TestBed#get

codeNoobie:
Cool. So I will add similar deprecation message in Spectator.

codeNoobie:
@NetanelBasal Added. Please check and let me know if something else is missing.

PR #267, opened 2020-02-13
chore: upgrading to angular v9

codeNoobie:
- Upgraded angular dependencies to v9
- Removed deprecated initUIEvent with UIEvent constructor
- With latest typescript, spyOnProperty getter test failed for es2015
  due to js compilation. Changed test target to es5 to fix this.
- Upgraded karma and jasmine to latest

BREAKING CHANGE: This PR sets the angular peer dependencies to v9

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[X] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, spectator is on Angular 8. It needs to be updated to v9, so that it can stay up to date with angular.
Issue Number: #262

## What is the new behavior?
Upgraded to Angular 9 and Typescript 3.7. Also made upgrades to test libraries.

```
[X] Yes
[ ] No
```

Since we are upgrading peer dependencies on Angular from v8 to v9, I am marking this as a breaking change. The consumers will have to upgrade to Angular v9 if they want to use this version of spectator.


jogelin:
I don't see any modifications related to  `entryComponents`, shouldn't we remove it ?

I had to rollback entryComponents on my tests using spectator but I could remove them in my code when migrating to angular 9

LayZeeDK:
This PR introduces breaking changes in terms of Angular and related dependencies so it requires a major release.

NetanelBasal:
Sure, it will be v5.

LayZeeDK:
@codeNoobie the type would then be `feat`, not `chore`.

NetanelBasal:
@jogelin, for now, I prefer leaving `entryComponents`. You can remove it if you are working with Ivy.

codeNoobie:
> It makes more sense to create a type such as:
> 
> `export type Token<T> = Type<T> | InjectionToken<T> | AbstractType<T>`
> 
> and use it in all places :)

The current Token<T> declaration uses locally declared `InjectableType<T>`. I am removing this as it will get covered by Type<T> and AbstractType<T> now.


```
/**
 * Type for classes that can be used as tokens for injection. We might be inclined to use Angular's `Type<T>` interface for this, however,
 * that interface cannot be used for abstract classes, which are also valid tokens. The `InjectableType<T>` supports both abstract and
 * concrete classes and thus is a more accurate definition of a class type which can be used for injection.
 */
export type InjectableType<T> = Function & { prototype: T };

/** Type representing valid typesafe token types for provider binding. */
export type Token<T> = InjectableType<T> | InjectionToken<T>;
```

NetanelBasal:
Thanks, I will test it locally asap, and if everything is ok I will merge it with #266

PR #281, opened 2020-02-28
Bugfix/jest tobevisible

### LINKED ISSUE #280: toBeVisible-matcher does not work with jest
## I'm submitting a...
<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior
Visible elements are reported as "not being visible" by the `toBeVisible`-matcher, when spectator is used within jest (using the `@ngneat/spectator/jest`dependency). 

## Expected behavior
Visible elements should be reported as "being visible"

## Minimal reproduction of the problem with instructions
A minimal reproduction repository can be found here: https://github.com/askarby/jest-tobevisible-bug

## What is the motivation / use case for changing the behavior?
You would expect the matcher to work, used in jest or jasmine alike.

## Environment

<pre><code>
Angular version: 9.0.3
</code></pre>

Browser:
<pre><code>
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
- [x] jsdom
</code></pre>
 
For Tooling issues:
- Node version: 10.17.0
- Platform:  Mac

## Additional Details
The problem seems to be related to the issue: https://github.com/jsdom/jsdom/issues/135 (regarding jsdom not being able to determine offsetHeight, offsetWidth, etc (which is used from within the `toBeVisible`-matcher).

askarby:
Fixes #280


projects/spectator/src/lib/matchers.ts
```diff
@@ -340,18 +340,29 @@ function isHidden(elOrSelector: HTMLElement | string): boolean {
...
 
-    if (
-      !(el.offsetWidth || el.offsetHeight || el.getClientRects().length) ||
-      el.style.display === 'none' ||
-      el.style.visibility === 'hidden' ||
-      el.type === 'hidden' ||
-      el.hasAttribute('hidden')
-    ) {
+    const or = (previous, current) => current || previous;
+    const elementIsHidden = hiddenWhen.map(rule => rule(el)).reduce(or, false);
```
NetanelBasal: We can use `elementIsHidden = hiddenWhen.some(Boolean)`

projects/spectator/src/lib/matchers.ts
```diff
@@ -340,18 +340,29 @@ function isHidden(elOrSelector: HTMLElement | string): boolean {
...
 
+  const hiddenWhen = [
+    el => !(el.offsetWidth || el.offsetHeight || el.getClientRects().length),
+    el => el.style.display === 'none',
+    el => el.style.visibility === 'hidden',
+    el => el.type === 'hidden',
+    el => el.hasAttribute('hidden')
+  ];
+
+  const runningInJsDom = navigator.userAgent.includes('Node.js') || navigator.userAgent.includes('jsdom');
```
NetanelBasal: Let's move it to an export function so that we re-use it in the future with other special cases.

README.md
```diff
@@ -943,6 +943,11 @@ expect('div').toHaveDescendant('.child');
 expect('div').toHaveDescendantWithText({selector: '.child', text: 'text'});
 ```
 
+Please notice that due to restrictions within Jest (not applying actual layout logic), certain matchers may give false positives - eg.:
```
NetanelBasal: Can we make it short and clear, please? I think one sentence is enough.

projects/spectator/src/lib/matchers.ts
askarby: Agreed!

projects/spectator/src/lib/matchers.ts
askarby: Agreed, this is most likely not the last time that JSDOM differs from the a browser DOM.

README.md
askarby: Agreed... not sure if my rewording is short (enough)?

Issue #283, opened 2020-03-03
Version 5.0.0 has BREAKING CHANGES and there no updated alternative branches


Coffee-Tea:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[x] Other... Please describe:
</code></pre>

## Current behavior

Version 5.0.0 has BREAKING CHANGES and requires Angular v.9 but  it's pretty recent version and I believe there are a lot of projects with less Angular version.
It's not a problem to use 4.11.1 version as the latest one before BREAKING CHANGES but **new fixes/updated included in new versions are not available** for most of projects with Angular less than 9.

## Expected behavior

Would be great to have separate branch with versioning 4.12.X, 14.13.X and so on for Angular versions before v9 and available fixes/updates (e.g. add in 5.1.0 etc.) on that branch too.

## Minimal reproduction of the problem with instructions
For bug reports please provide the *STEPS TO REPRODUCE* and if possible a *MINIMAL DEMO* of the problem via
https://stackblitz.com or similar (you can use this template as a starting point: https://bit.ly/2zme3bj).

## What is the motivation / use case for changing the behaviour?

There are a lot of projects with less Angular version than v9 and they can't use the latest fixes/updates which are available only for projects with Angular v.9 b/c are merged to master branch after 5.0.0 version.

## Environment

<pre><code>
Angular version: 7.2.0

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  10.16.2
- Platform:  Mac

Others:

</code></pre>

NetanelBasal:
Truthfully, I've thought about this a lot, and I agree. I believe that we can downgrade the peer dependency, and add protection in the new `spectator.inject()` method. Do you want to try it, and create a PR?

Coffee-Tea:
@NetanelBasal thanks, I will take a look

PR #284, opened 2020-03-03
feat(lib): use TestBed.get if .inject falsy - back compatibility (#283)

Coffee-Tea:
## PR Checklist

## PR Type
Backward compatibility change.
Allow to use 5.x lib version for Angular version less than v9.
Keep to use for Angular v9 as is.

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #283 

## What is the new behavior?
You can continue using the latest version of Spectator in you Angular projects with version less than 9 (e.g. 7, 8 etc.)

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/spectator/package.json
+++ b/projects/spectator/package.json
@@ -29,9 +29,9 @@
     "replace-in-file": "^4.1.3"
   },
   "peerDependencies": {
-    "@angular/common": "^8.0.0",
-    "@angular/core": "^8.0.0",
-    "@angular/router": "^8.0.0",
+    "@angular/common": ">= 8.0.0",
+    "@angular/core": ">= 8.0.0",
+    "@angular/router": ">= 8.0.0",
     "typescript": ">= 3.7.0"
   },
   "bin": {
--- a/projects/spectator/package.json
+++ b/projects/spectator/package.json
@@ -32,7 +32,7 @@
     "@angular/common": ">= 8.0.0",
     "@angular/core": ">= 8.0.0",
     "@angular/router": ">= 8.0.0",
-    "typescript": ">= 3.7.0"
+    "typescript": ">= 2.8.0"
   },
   "bin": {
     "spectator-migrate": "./migrate.js"
--- a/projects/spectator/package.json
+++ b/projects/spectator/package.json
@@ -29,9 +29,9 @@
     "replace-in-file": "^4.1.3"
   },
   "peerDependencies": {
-    "@angular/common": "^9.0.1",
-    "@angular/core": "^9.0.1",
-    "@angular/router": "^9.0.1",
+    "@angular/common": "^8.0.0",
+    "@angular/core": "^8.0.0",
+    "@angular/router": "^8.0.0",
     "typescript": ">= 3.7.0"
   },
   "bin": {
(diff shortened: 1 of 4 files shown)


NetanelBasal:
You need to update the `[peerDependencies](https://github.com/ngneat/spectator/blob/master/projects/spectator/package.json#L31)`.

projects/spectator/package.json
```diff
@@ -29,9 +29,9 @@
     "replace-in-file": "^4.1.3"
   },
   "peerDependencies": {
-    "@angular/common": "^9.0.1",
-    "@angular/core": "^9.0.1",
-    "@angular/router": "^9.0.1",
+    "@angular/common": "^8.0.0",
```
NetanelBasal: This doesn't include v9. You need to use a range.

projects/spectator/package.json
```diff
@@ -29,9 +29,9 @@
     "replace-in-file": "^4.1.3"
   },
   "peerDependencies": {
-    "@angular/common": "^9.0.1",
-    "@angular/core": "^9.0.1",
-    "@angular/router": "^9.0.1",
+    "@angular/common": ">= 8.0.0",
+    "@angular/core": ">= 8.0.0",
+    "@angular/router": ">= 8.0.0",
     "typescript": ">= 3.7.0"
```
NetanelBasal: "typescript": ">= 2.8.0"

Coffee-Tea: let me check with this version of typescript,
I haven't checked it and not sure there are no new typescript features that are used in code which will not work with 2.8.0.

I will update you

Coffee-Tea: @NetanelBasal I would put "typescript": ">= 3.5.3" or smth like that b/c there are a bunch of error when using 2.8.0. Actually, 3.5.3 was used as dependency in the root package.json:
https://github.com/ngneat/spectator/commit/e13c9554778bdb179dfc7235aedb4b3b90302850#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L77

Here is a part of errors I'm getting with 2.8.0:
image


Please confirm

NetanelBasal: This was the lastest version before we release v5: https://github.com/ngneat/spectator/blob/5c98661cb61fe684f09b915b17629ca37749f867/projects/spectator/package.json#L35

Coffee-Tea: @NetanelBasal yes you're right, but what is goal of peerDependencies? Maybe I'm wrong but should not we specify versions which gonna work with other specified peerDependencies versions?
I assume that it was wrong with >= 2.8.0 but if you're ok I will get back to it.

NetanelBasal: > Here is a part of errors I'm getting with 2.8.0:

What is the Angular version you're using?

Coffee-Tea: Let me check 2.8.0 with 8.0.0

Coffee-Tea: Doesn't work for me,
also 8.0.0 angular is using 3.4.2 version https://github.com/angular/angular/blob/38a7e2a775d2c3495a8574dde7d1a7106dd0ae04/package.json#L108

so I'm really unsure in 2.8.0 but pushed it. It doesn't break anything, just seems to be smth like disinformation.

NetanelBasal: The TS version depends on your Angular version.

Coffee-Tea: Yes, that what I mean, and I'm not sure that TS 2.8.0 is compatible with Angular 8.0.0 :)

PR #295, opened 2020-04-17
feat(dom-selectors): add `byTextContent` query

th0r:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: I've also updated Prettier to v1.19.1 in `package-lock.json` (this version still matches the version from `package.json`) because old version didn't support optional chaining operator which I use in this PR.
```

Current `byText` query matches text only in direct element's text nodes and doesn't support nested elements.

For example, `byText('foobar')` won't match `div` in this example:
```html
<div>
  <span>foo</span><span>bar</span>
</div>
```

Issue Number: N/A

## What is the new behavior?

This PR introduces `byTextContent` query that eliminates this restriction. I've made `selector` option mandatory because queries like `byTextContent('foo', {exact: false})` will match the element with text `foo` and all its parents which, I think, will be very confusing is not good.

```
[ ] Yes
[x] No
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,19 @@ spectator.query(byAltText('By alt text'));
 spectator.query(byLabel('By label'));
 spectator.query(byText('By text'));
 spectator.query(byText('By text', {selector: '#some .selector'}));
+spectator.query(byTextContent('By text content', {selector: '#some .selector'}));
 ```
+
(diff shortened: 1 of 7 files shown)


NetanelBasal:
LGTM

th0r:
@NetanelBasal added docs

Issue #337, opened 2020-08-18
Overriding Injection Tokens not working for createComponentFactory


Kurtz1993:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior
I am trying to override a provider's value that's tied to an InjectionToken with no luck.

This is my createComponentFactory function where I'm setting up the providers.
![image](https://user-images.githubusercontent.com/5412470/90542115-3f5ad100-e149-11ea-973d-3b35b8300a8f.png)

Here, in my test where I'm trying to override that same provider.
![image](https://user-images.githubusercontent.com/5412470/90542152-4c77c000-e149-11ea-9f50-a420a3e68b1d.png)

This is the result. Nothing had changed.
![image](https://user-images.githubusercontent.com/5412470/90542182-55689180-e149-11ea-972d-f117e3255a2d.png)

## Expected behavior
The provider should be overriden.

## Environment

<pre><code>
Angular version: 10.0.8

Browser:
- [ x ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 10.20.1
- Platform:  Windows 10

Others:

</code></pre>

NetanelBasal:
Please provide a reproduction example.

Kurtz1993:
@NetanelBasal here's a quick example of that.

https://github.com/Kurtz1993/spectator-issue

adrianrutkowski:
@Kurtz1993 in this test case

`it('should get the mocked instance of STATE injection token', () => { ... })`

you are trying to recreate component when TestBed is already initialized under the hood by

`beforeEach(() => {
    spectator = createComponent();
  });`

It's an inconsistency on the Angular side, because it throws error when using override... methods after TestBest initialization, but not in overrideProvider case. I have created pull request which should fix this issue in Angular repository https://github.com/angular/angular/pull/38717

bboydflo:
[...]

PR #383, opened 2021-01-23
fix: override providers for createComponentFactory

stefanoww:
## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Overriding providers after executing createComponentFactory is impossible.

Issue Number: N/A
#337

## What is the new behavior?
When user want to override providers, testing module is reinitialized with new providers.

```
[ ] Yes
[X] No
```


stefanoww:
@NetanelBasal I am not sure whether that is the best approach. From that what I was able to figure out, we can not override providers after calling compileComponents. If we want to override it, we need to complete reset current configuration and create new one. That's what I have done.

I think similar problem may occur within other create-factory.ts files. If current solution will be fine I may also fix it in other places.

stefanoww:
@NetanelBasal any update with this one?

NetanelBasal:
I don't think we need to use `resetTestingModule` for every provider. The issue is only when using `InjectionToken`, right?

stefanoww:
To be honest based one the that answer from issue https://github.com/ngneat/spectator/issues/337#issuecomment-689657302 and linked there issue from Angular repo I am not sure that problem occur only for **InjectionToken**. I have understood that you just can not override providers (any) when TestBed is already instantiated.

NetanelBasal:
Now I'm not sure what's the issue. Works for me - https://github.com/ngneat/bind-query-params/blob/master/projects/ngneat/bind-query-params/src/lib/lib.spec.ts#L115

stefanoww:
Maybe to better figure out that issue you may get into my branch and play with code and tests. What I understand as the problem there is that  we just cannot override providers when the TestBed.compileComponents() is called and that what's mostly happen in beforeach block where we execute createComponentFactory.

PR #402, opened 2021-03-03
Update package peerDependencies

### LINKED ISSUE #393: Document need for Angular 11 because of waitForAsync function
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[X] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

When using Spector 7.0.0 with Angular version < 11 you get an error about waitForAsync not found.  

## Expected behavior

Does the package.jason file need to be updated to document that the Angular version should be >= 11>

## Minimal reproduction of the problem with instructions
For bug reports please provide the *STEPS TO REPRODUCE* and if possible a *MINIMAL DEMO* of the problem via
https://stackblitz.com or similar (you can use this template as a starting point: https://bit.ly/2zme3bj).
Create an Angular project using Angular version 10, 9 or 8.  Try to run Spectator.  

## What is the motivation / use case for changing the behavior?

It took some time to track down the change in Specator 7.0.0 that caused the error.  

## Environment

<pre><code>
Angular version: 10.0.4 puls 9 and 8

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

matheo:
Update package peerDependencies and also the project dependencies. Fixes #393


matheo:
@NetanelBasal WDYT of syncing `spectator` version with the `@angular` one?
I see that many projects avoid this problem using that convention.

NetanelBasal:
I think the waitForAsync is from v10.1, no? This is the only breaking change?

matheo:
ah right, it was available in `v10.1`

> This is the only breaking change?
Yeah, the schematic code is available at `v10.0`

PR #405, opened 2021-03-05
build: update package dependencies

matheo:
Hmmm well, this is the reason I'm adding `lerna` + `yarn workspaces` to my NX and NG CLI repos...
the dependency management gets easier. WDYT of giving it a try @NetanelBasal?

### DIFF
--- a/projects/spectator/package.json
+++ b/projects/spectator/package.json
@@ -23,9 +23,9 @@
     "angular testing dumb components"
   ],
   "dependencies": {
-    "@testing-library/dom": "6.1.0",
-    "jquery": "3.5.0",
-    "replace-in-file": "^4.1.3",
+    "@testing-library/dom": "7.26.5",
+    "jquery": "3.6.0",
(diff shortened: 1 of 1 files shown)


matheo:
Maybe is enough with `yarn workspaces`

NetanelBasal:
I'm open to suggestions

projects/spectator/package.json
```diff
@@ -23,9 +23,9 @@
     "angular testing dumb components"
   ],
   "dependencies": {
-    "@testing-library/dom": "6.1.0",
-    "jquery": "3.5.0",
-    "replace-in-file": "^4.1.3",
+    "@testing-library/dom": "7.26.5",
```
NetanelBasal, line 26: Are there any breaking changes in these packages?

matheo: No breaking changes, `v7.0.1` was built and is already using these versions, and I've installed it without issues so they are compatible

Issue #414, opened 2021-03-15
TriggerEventHandler should provide an option to query from 'root'


anandtiwary:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

Some components create popover elements which are part of the Dom root. These dom elements are not descendants of the spectator host element. The spectator api provides an option to query them using the `{root: true}` option. This works well.

With my use case, I am trying to trigger an event handler using `triggerEventHandler` method on a child component which is being displayed inside the  popover. This of course doesn't work. I think we would need a similar `{root: true}` option to support this use case. If there is an alternative could you please point me towards the same?

## Expected behavior

We should be able to `triggerEventHandler` method with the root option.

## Minimal reproduction of the problem with instructions
For bug reports please provide the *STEPS TO REPRODUCE* and if possible a *MINIMAL DEMO* of the problem via
https://stackblitz.com or similar (you can use this template as a starting point: https://bit.ly/2zme3bj).

1. Create a host component that has a Popover component inside (or any thing that attaches the content to the root dom). 
2. Add any other component to this popover's content. 
3. From the host component, trigger the popover to show its content in an overlay
4. Now Trigger an event handler using `triggerEventHandler` on the **child component**. This should trigger corresponding handler in the host component

## What is the motivation / use case for changing the behavior?

Missing behavior

## Environment

<pre><code>
Angular version: 11.X

Browser:
- [X] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX   v15.3.0
- Platform:   MAc

Others:

</code></pre>

NetanelBasal:
You're welcome to submit a PR.

PR #416, opened 2021-03-17
Trigger event handler

anandtiwary:
## PR Checklist
fix: adding a root option to trigger event handler method

TriggerEventHandler method currently doesn't support a search from root. Some angular components can create overlay elements which are part of the root and so, are not descendants of the current test dom element.

- [X] Docs have been added / updated (for bug fixes / features)

## PR Type
This change adds the support for finding the trigger component from the root. For this we are traversing up the DebugElement tree using the parent property. It is still a bounded search since the tree traversal will end at the root. Happy to explore other better approaches.

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngneat/spectator/issues/414

## What is the new behavior?
Adds a new option to triggerEventHandler method that would allow searching for element from root, similar to query method.

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/spectator/package.json
+++ b/projects/spectator/package.json
@@ -32,7 +32,6 @@
     "@angular/common": ">= 11.1.0",
     "@angular/core": ">= 11.1.0",
     "@angular/router": ">= 11.1.0",
-    "@angular/cdk": ">= 11.1.0",
     "typescript": ">= 2.8.0"
   },
   "bin": {
--- a/projects/spectator/package.json
+++ b/projects/spectator/package.json
@@ -29,9 +29,9 @@
     "tslib": "^2.0.0"
   },
   "peerDependencies": {
-    "@angular/common": ">= 11.1.0",
-    "@angular/core": ">= 11.1.0",
-    "@angular/router": ">= 11.1.0",
+    "@angular/common": ">= 10.1.0",
+    "@angular/core": ">= 10.1.0",
+    "@angular/router": ">= 10.1.0",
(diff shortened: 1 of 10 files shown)


projects/spectator/package.json
```diff
@@ -29,9 +29,10 @@
     "tslib": "^2.0.0"
   },
   "peerDependencies": {
-    "@angular/common": ">= 10.1.0",
-    "@angular/core": ">= 10.1.0",
-    "@angular/router": ">= 10.1.0",
+    "@angular/common": ">= 11.1.0",
+    "@angular/core": ">= 11.1.0",
+    "@angular/router": ">= 11.1.0",
+    "@angular/cdk": ">= 11.1.0",
```
NetanelBasal, line 35: We don't need to cdk here.

anandtiwary: Yeah this is not needed. Do you want to keep the other peer dep bump?

projects/spectator/package.json
```diff
@@ -29,9 +29,9 @@
     "tslib": "^2.0.0"
   },
   "peerDependencies": {
-    "@angular/common": ">= 10.1.0",
-    "@angular/core": ">= 10.1.0",
-    "@angular/router": ">= 10.1.0",
+    "@angular/common": ">= 11.1.0",
```
NetanelBasal, line 32: Please revert

Issue #361, opened 2020-10-24
About toHaveClass


dev054:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[?] Bug report  
[ ] Performance issue
[?] Feature request
[ ] Documentation issue or request
[ ] Support request
[?] Other... Please describe:

I don't know if it's the expected behavior or a bug.
</code></pre>

## Current behavior
I have this template:

```html
<div
  class="primary"
  [class.test]="someCondition"
  [class.awesome]="anotherCondition"
>A text</div>
```

and this spec:
```ts
expect(spectator.query(byText('A text'))).toHaveClass([
  'primary',
  'awesome',
  'test',
]);
```

...the problem is that the custom matcher `toHaveClass` expects the array to be ordered as in the same way it's rendered by Angular so to make the tests green I have to look at its rendering order to replicate in spec and I don't know if it's the expected behavior.

## Expected behavior
If it's the expected behavior, I'd suggest to create an `options` parameter to make it configurable as people may want only check the classes without caring about its order.

## Environment

<pre><code>
Angular version: 10

NetanelBasal:
You're welcome to create a PR.

ntvsx193:
The jasmine has support native `toHaveClass(expected: string, expectationFailOutput?: any): boolean;` from version v.3.0.0.
@dev054 it's have breaking changes with this naming.
May be name for `toHaveClassSerial` ... or any some other name.

PR #424, opened 2021-04-02
Feature/fix: toHaveClass should accept classes in any order if configured

YoeriNijs:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?
This is a proposal to add additional support for the following by passing an optional config object: https://github.com/ngneat/spectator/issues/361

```
[] Yes
[x ] No
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1079,9 +1079,6 @@ The same rules also apply to directives using the `directiveProviders` and `dire
 expect('.zippy__content').not.toExist();
 expect('.zippy__content').toHaveLength(3);
 expect('.zippy__content').toHaveId('id');
-expect('.zippy__content').toHaveClass('class');
-expect('.zippy__content').toHaveClass('class a, class b');
-expect('.zippy__content').toHaveClass(['class a', 'class b']);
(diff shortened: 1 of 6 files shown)


NetanelBasal:
Thanks for the PR. I prefer to add a configurable option to `toHaveClass`:

```ts
toHaveClass('', { strict: false })
```

YoeriNijs:
Yes, done! I was doubting a bit whether we should pass a config or just a strict boolean, but I guess I prefer the config since we could extend it with more ease eventually (if needed).

Now, our toHaveClass check is not strict anymore. I guess this is not a problem since it was unintended anyway. Agree?

NetanelBasal:
Unfortunately, we can't count on it and break the code for some users. We should make the default `strict: true` for now, and change it to `false` in the next major version.

YoeriNijs:
Done! :)

Issue #373, opened 2020-12-04
SpectatorHost.focus() is broken in a few ways when using Jest


johncrim:
## I'm submitting a...
<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior
When using jest, calls to `SpectatorHost.focus()` do dispatch a (fake) `focus` event, but the current implementation prevents automatic `blur` events from occurring, and also causes the `toBeFocused()` matcher to not work. This is due to the call to `patchElementFocus()` in `DomSpectator.focus()`, which overwrites valid and more useful jsdom `focus()` methods with versions that don't work as well.

The following test shows the issue, and a workaround:

```ts
import { ChangeDetectionStrategy, Component } from '@angular/core';
import { createHostFactory, SpectatorHost } from '@ngneat/spectator/jest';

@Component({
  selector: 'ui-test-focus',
  template: `<button id="button1" (focus)="countFocus('button1')" (blur)="countBlur('button1')">Button1</button>
             <button id="button2" (focus)="countFocus('button2')" (blur)="countBlur('button2')">Button2</button>`,
  changeDetection: ChangeDetectionStrategy.OnPush,
  host: {
    '[attr.tabindex]': '0',
    '(focus)': 'countFocus("ui-test-focus")',
    '(blur)': 'countBlur("ui-test-focus")'
  }
})
export class TestFocusComponent {

  private readonly focusCounts = new Map<string, number>();
  private readonly blurCounts = new Map<string, number>();

  public countFocus(id: string) {
    this.focusCounts.set(id, this.focusCount(id) + 1);
  }

  public countBlur(id: string) {
    this.blurCounts.set(id, this.blurCount(id) + 1);
  }

  public focusCount(id: string): number {
    return this.focusCounts.get(id) ?? 0;
  }

  public blurCount(id: string): number {
    return this.blurCounts.get(id) ?? 0;
  }

}

describe('Spectator focus() in jest', () => {

  const createHost = createHostFactory(TestFocusComponent);
  let host: SpectatorHost<TestFocusComponent>;

  beforeEach(() => {
    host = createHost('<ui-test-focus></ui-test-focus>');
  })

  it('SpectatorHost.focus() in jest does not track active element', () => {
    host.focus('#button1');

    // FAILS
[...]

PR #473, opened 2021-08-08
Bug/373 jest focus

johncrim:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #373

## What is the new behavior?
1. In Jest (more specifically jsdom), the patch for `HTMLElement.focus()` and `HTMLElement.blur()` is not applied.
2. In browsers, when the  `HTMLElement.focus()` is patched the patched implementation is more robust. It calls the browser's `HTMLElement.focus()` method first, and if the focus and/or blur events are not detected when `focus()` is called, it sends the fake events. This ensures that `document.activeElement` is set (so the `toBeFocused()` matcher works, and app behavior that depends on `document.activeElement` works), and it ensures that a single focus event, and 0-1 blur events, are emitted.

```
[ ] Yes
[x] No
```

There is no API change, but there is a change in behavior, that could conceivably break tests that depend on the old behavior (most likely to work around it). However this should be unlikely.

`focus()` and `blur()` were functionally broken in Jest, so this shouldn't break jest tests. The change for jest is that the jsdom focus() and blur() methods are no longer patched, but they didn't need to be.

With the previous implementation, `focus()` and `blur()` worked partially in karma, but `document.activeElement` wasn't set by the patched methods. The old behavior ensured that focus and blur events were always sent when their respective methods were called, but it didn't handle blurring the previous activeElement. It also didn't set the activeElement. The new implementation calls the browser's `focus()` method, which seems to always set the `document.activeElement`, but doesn't always send focus/blur events depending on the browser state. The new patch listens for browser-emitted focus/blur events, and sends fake ones (same as before) if they weren't emitted by the browser.

It might be a good idea to call the browser's `blur()` method first, and send a blur event only if one doesn't occur. However I haven't had to deal with this possible inconsistency between test environments and browsers, so I didn't make this fix.

### DIFF
--- a/projects/spectator/src/lib/internals/element-focus.ts
+++ b/projects/spectator/src/lib/internals/element-focus.ts
@@ -4,8 +4,47 @@ import { isRunningInJsDom } from '../utils';
 /** Property name added to HTML Elements to ensure we don't double-patch focus methods on an element. */
 const IS_FOCUS_PATCHED_PROP = '_patched_focus';
 
+/** Ensures that a single set of matching focus and blur events occur when HTMLElement.focus() is called. */
+class FocusEventWatcher implements EventListenerObject {
+
+  private readonly _active: Element | null;
+  private _blurred = false;
+  private _focused = false;
+
+  constructor(private readonly _e: HTMLElement) {
+    this._active = _e.ownerDocument.activeElement;
+    this._e.addEventListener('focus', this);
+    this._active?.addEventListener('blur', this);
+  }
+  public handleEvent(evt: Event): void {
+    if (evt.type === 'focus') {
+      this._focused = true;
+    }
+    else if (evt.type === 'blur') {
+      this._blurred = true;
+    }
+  }
+
+  /**
+   * If focus and blur events haven't occurred, fire fake ones.
+   */
+  public ensureFocusEvents() {
+    this._e.removeEventListener('focus', this);
+    this._active?.removeEventListener('blur', this);
+
+    // Ensure activeElement is blurred
+    if (this._active && !this._blurred && this._active === this._e.ownerDocument.activeElement) {
+      dispatchFakeEvent(this._active, 'blur');
+    }
+
+    if (!this._focused) {
+      dispatchFakeEvent(this._e, 'focus'); // Needed to cause focus event
+    }
+  }
+}
+
 /**
- * Patches an elements focus and blur methods to emit events consistently and predictably.
+ * Patches an element's focus and blur methods to emit events consistently and predictably in tests.
  * This is necessary, because some browsers, like IE11, will call the focus handlers asynchronously,
  * while others won't fire them at all if the browser window is not focused.
  *
@@ -17,15 +56,12 @@ export function patchElementFocus(element: HTMLElement): void {
   if (!isRunningInJsDom() && (element[IS_FOCUS_PATCHED_PROP] === undefined)) {
     const baseFocus = element.focus.bind(element);
     element.focus = (options) => {
-      // Blur current active
-      const active = element.ownerDocument.activeElement;
-      if (active) {
-        dispatchFakeEvent(active, 'blur');
-      }
+      const w = new FocusEventWatcher(element);
 
-      // Focus
-      baseFocus(options); // Needed to set document.activeElement
-      dispatchFakeEvent(element, 'focus'); // Needed to cause focus event
+      // Sets document.activeElement. May or may not send focus + blur events
+      baseFocus(options);
+
+      w.ensureFocusEvents();
     }
     element.blur = () => dispatchFakeEvent(element, 'blur');
     element[IS_FOCUS_PATCHED_PROP] = true;
--- a/projects/spectator/src/lib/internals/element-focus.ts
+++ b/projects/spectator/src/lib/internals/element-focus.ts
@@ -1,4 +1,8 @@
 import { dispatchFakeEvent } from '../dispatch-events';
+import { isRunningInJsDom } from '../utils';
+
+/** Property name added to HTML Elements to ensure we don't double-patch focus methods on an element. */
+const IS_FOCUS_PATCHED_PROP = '_patched_focus';
 
 /**
  * Patches an elements focus and blur methods to emit events consistently and predictably.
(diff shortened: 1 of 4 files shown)


johncrim:
Let me know if the docs need to be updated - I'm not aware of a need.

projects/spectator/test/focus/test-focus.component.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    expect(host.component.focusCount('app-test-focus')).toBe(1);
+    expect(host.component.blurCount('app-test-focus')).toBe(1);
+    expect(host.component.focusCount('button1')).toBe(1);
+    expect(host.component.blurCount('button1')).toBe(1);
+    expect(host.component.focusCount('button2')).toBe(1);
+    expect(host.component.blurCount('button2')).toBe(0);
+  });
+
+  it('calling focus() multiple times does not cause multiple patches', () => {
```
NetanelBasal, line 32: Could you please add this spec to Jest?

projects/spectator/src/lib/internals/element-focus.ts
```diff
@@ -1,13 +1,69 @@
 import { dispatchFakeEvent } from '../dispatch-events';
+import { isRunningInJsDom } from '../utils';
+
+/** Property name added to HTML Elements to ensure we don't double-patch focus methods on an element. */
+const IS_FOCUS_PATCHED_PROP = '_patched_focus';
```
NetanelBasal, line 5: We can use Symbol

projects/spectator/src/lib/internals/element-focus.ts
```diff
@@ -1,13 +1,69 @@
...
+
+/** Ensures that a single set of matching focus and blur events occur when HTMLElement.focus() is called. */
+class FocusEventWatcher implements EventListenerObject {
+
+  private readonly _active: Element | null;
+  private _blurred = false;
+  private _focused = false;
+
+  constructor(private readonly _e: HTMLElement) {
```
NetanelBasal, line 14:
```suggestion
  constructor(private readonly element: HTMLElement) {
```

projects/spectator/src/lib/internals/element-focus.ts
```diff
@@ -1,13 +1,69 @@
...
+    this._active = _e.ownerDocument.activeElement;
+    this._e.addEventListener('focus', this);
+    this._active?.addEventListener('blur', this);
+  }
+  public handleEvent(evt: Event): void {
```
NetanelBasal, line 19:
```suggestion
  public handleEvent({ type }: Event): void {
```

projects/spectator/src/lib/internals/element-focus.ts
```diff
@@ -1,13 +1,69 @@
...
+    if (evt.type === 'focus') {
```
NetanelBasal, line 20:
```ts
this._focused = type === 'focus';
this._blurred = type === 'blur';
```

projects/spectator/src/lib/internals/element-focus.ts
```diff
@@ -1,13 +1,69 @@
...
  */
 export function patchElementFocus(element: HTMLElement): void {
-  element.focus = () => dispatchFakeEvent(element, 'focus');
-  element.blur = () => dispatchFakeEvent(element, 'blur');
+
+  // https://github.com/ngneat/spectator/issues/373 - Don't patch when using JSDOM, eg in Jest
+  if (!isRunningInJsDom() && (element[IS_FOCUS_PATCHED_PROP] === undefined)) {
+    const baseFocus = element.focus.bind(element);
+    element.focus = (options) => {
+      const w = new FocusEventWatcher(element);
```
NetanelBasal, line 59:
```suggestion
      const focusEventWatcher = new FocusEventWatcher(element);
```

projects/spectator/src/lib/internals/element-focus.ts
```diff
@@ -1,13 +1,69 @@
...
  *
  * patchElementFocus(triggerEl);
  */
 export function patchElementFocus(element: HTMLElement): void {
-  element.focus = () => dispatchFakeEvent(element, 'focus');
-  element.blur = () => dispatchFakeEvent(element, 'blur');
+
+  // https://github.com/ngneat/spectator/issues/373 - Don't patch when using JSDOM, eg in Jest
+  if (!isRunningInJsDom() && (element[IS_FOCUS_PATCHED_PROP] === undefined)) {
+    const baseFocus = element.focus.bind(element);
```
NetanelBasal, line 57:
```suggestion
    const originalFocus = element.focus.bind(element);
```

projects/spectator/src/lib/internals/element-focus.ts
```diff
@@ -1,13 +1,69 @@
...
+
+/** Ensures that a single set of matching focus and blur events occur when HTMLElement.focus() is called. */
+class FocusEventWatcher implements EventListenerObject {
+
+  private readonly _active: Element | null;
+  private _blurred = false;
+  private _focused = false;
+
+  constructor(private readonly _e: HTMLElement) {
+    this._active = _e.ownerDocument.activeElement;
```
NetanelBasal, line 15:
```suggestion
    this.activeElement = _e.ownerDocument.activeElement;
```

projects/spectator/src/lib/internals/element-focus.ts
```diff
@@ -1,13 +1,69 @@
...
+
+  /**
+   * If focus and blur events haven't occurred, fire fake ones.
+   */
+  public ensureFocusEvents() {
+    this._e.removeEventListener('focus', this);
+    this._active?.removeEventListener('blur', this);
+
+    // Ensure activeElement is blurred
+    if (this._active && !this._blurred && this._active === this._e.ownerDocument.activeElement) {
```
NetanelBasal, line 36: Isn't the following condition more accurate? 
```ts
if(!this._blurred && this.activeElement === this.element) {}
```

NetanelBasal (changes requested):

projects/spectator/src/lib/internals/element-focus.ts
```diff
@@ -1,13 +1,69 @@
...
 /**
- * Patches an elements focus and blur methods to emit events consistently and predictably.
+ * Patches an element's focus and blur methods to emit events consistently and predictably in tests.
  * This is necessary, because some browsers, like IE11, will call the focus handlers asynchronously,
  * while others won't fire them at all if the browser window is not focused.
  *
  * patchElementFocus(triggerEl);
  */
 export function patchElementFocus(element: HTMLElement): void {
-  element.focus = () => dispatchFakeEvent(element, 'focus');
```
NetanelBasal, line 11: Why deleting the Jaminse functionality?

projects/spectator/test/focus/test-focus.component.spec.ts
johncrim: Sure.

projects/spectator/src/lib/internals/element-focus.ts
johncrim: That's a good idea.

projects/spectator/src/lib/internals/element-focus.ts
johncrim: This check is to make sure that the `document.activeElement` still matches the `document.activeElement` that was recorded when the watcher was created (before the browser's `element.focus()` was called). It could conceivably change, and I don't think we'd want the initial `activeElement` to receive a blur event if the `activeElement` has since changed.

projects/spectator/src/lib/internals/element-focus.ts
johncrim: It's replaced with a better patched focus().

projects/spectator/src/lib/internals/element-focus.ts
johncrim: That's incorrect - we are listening for both a blur event and a focus event on different elements, that would overwrite the first event when the second event occurs.

projects/spectator/src/lib/internals/element-focus.ts
johncrim: I went with `priorActiveElement`.

johncrim:
I've addressed all your feedback.

Issue #487, opened 2021-09-27
"Can't bind to 'queryParams' since it isn't a known property of 'a'" despite using createRoutingFactory


AnthonyLenglet:
### Discussed in https://github.com/ngneat/spectator/discussions/486

<div type='discussions-op-text'>

<sup>Originally posted by **AnthonyLenglet** September 27, 2021</sup>
Hi !

I've been using spectator to add tests to one of our existing project and I'm currently stumbling on an error related to `queryParams` in this component:

```html
<article class="wrapper">
  <header>
    <nav>
      <ul>
        <li mat-ripple routerLinkActive="active">
          <a routerLink="." [queryParams]="{ pipeline: 'todo' }">
            To do
          </a>
        </li>
        <li mat-ripple routerLinkActive="active">
          <a routerLink="." [queryParams]="{ pipeline: 'under-review' }">
            Under Review
          </a>
        </li>
        <li mat-ripple routerLinkActive="active">
          <a routerLink="." [queryParams]="{ pipeline: 'reviewed' }">
            Reviewed
          </a>
        </li>
      </ul>
    </nav>
  </header>
</article>
```

In my tests I am creating it using `createRoutingFactory` in order to add the routing functionality (there is a data property in the factory option, but I removed it here among other things in order to keep this to the point):

```ts
  let spectator: Spectator<ViewComponent>;
  const createComponent = createRoutingFactory({
    component: ViewComponent,
    queryParams: {
      pipeline: 'todo',
    },
  });

  beforeEach(() => (spectator = createComponent()));

  it('should create', () => {
    expect(spectator.component).toBeTruthy();
  });
```

Here is the output I am getting:

```
✔ Browser application bundle generation complete.
ERROR: 'NG0303: Can't bind to 'queryParams' since it isn't a known property of 'a'.'
Chrome Headless 93.0.4577.82 (Mac OS 10.15.7): Executed 24 of 52 SUCCESS (0 secs / 0.41 secs)
ERROR: 'NG0303: Can't bind to 'queryParams' since it isn't a known property of 'a'.'
Chrome Headless 93.0.4577.82 (Mac OS 10.15.7): Executed 24 of 52 SUCCESS (0 secs / 0.41 secs)
ERROR: 'NG0303: Can't bind to 'queryParams' since it isn't a known property of 'a'.'
Chrome Headless 93.0.4577.82 (Mac OS 10.15.7): Executed 24 of 52 SUCCESS (0 secs / 0.41 secs)
[...]

PR #488, opened 2021-09-27
feat(Router): Import RouterTestingModule when stubs are enabled

AnthonyLenglet:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #487 

## What is the new behavior?

`createRoutingFactory` now imports RouterTestingModule when stubs are enabled, then adds the necessary stubs and overrides, thus adding the missing inputs and methods from router related features

```
[ ] Yes
[x] No
```

`createRoutingFactory` had an undocumented `mockRouterLink` that has been removed as it is no longer needed


projects/spectator/src/lib/spectator-routing/router-link-stub.ts
```diff
@@ -14,15 +14,32 @@ import { RouterLink } from '@angular/router';
 })
 export class RouterLinkDirectiveStub {
```
NetanelBasal, line 15: Thanks for the PR. Now that I'm thinking about it, I can't remember why we have a stub here. Can you try to use the original directive, please?

AnthonyLenglet: from what I can gather, it looks like I can't just do `moduleMetadata.declarations.push(RouterLink);` instead of `moduleMetadata.declarations.push(RouterLinkStub);`

since angular 9, you need to import the module of the library directly
https://github.com/angular/angular/issues/37047#issuecomment-634226777

so turns out I need to import `RouterTestingModule` along with the other stubs when they are enabled

I was able to make it work but I needed to mock the serializeUrl function in the routerStub, as angular was trying to parse the url using it internally
```
serializeUrl(): string {
  return '/';
},
```

AnthonyLenglet: I'm actually not sure if the routerStub is still needed since I'm now importing `RouterTestingModule`, it looks like it should really just stub `ActiveRoute`

AnthonyLenglet:
Just updated the initial PR post with a new summary
 
`createRoutingFactory` now imports RouterTestingModule when stubs are enabled, then adds the necessary stubs and overrides, thus adding the missing inputs and methods from router related features

NetanelBasal:
@AnthonyLenglet tests are failing.

PR #499, opened 2021-10-22
feat: allow providing `ModuleTeardownOptions`

arturovt:


selangley-wa:
@arturovt 
Since we are on Angular 13 now, should the newest versions of Spectator now default `destroyAfterEach` to `true`?

Issue #519, opened 2021-12-11
Runtime error on Angular 13.1.0


ChaseOxide:
### Is this a regression?

No

### Description

I've just upgraded from Angular 13.0.3 to 13.1.0 and started receiving this error.
I've tried starting a new Angular project with `ng new`, get spectator installed and have `@ngneat/spectator` imported in the spec file. The test runs without runtime error until I have it imported

### Please provide a link to a minimal reproduction of the bug

https://github.com/ChaseOxide/spectator-crash

### Please provide the exception or error you saw

```true
An error was thrown in afterAll
  Uncaught ReferenceError: global is not defined
  ReferenceError: global is not defined
      at Object.74757 (http://localhost:9876/_karma_webpack_/webpack:/node_modules/pretty-format/build/plugins/AsymmetricMatcher.js:10:1)
```

### Please provide the environment you discovered this bug in

```true
Angular 13.1.0
Spectator 9.0.0
Jasmine 3.10.0
```

### Anything else?

Output from `npm ls pretty-format`:
```
`-- @ngneat/spectator@9.0.0
  `-- @testing-library/dom@7.26.5
    `-- pretty-format@26.6.2 
```

### Do you want to create a pull request?

No

molily:
Bumping the @testing-library/dom dependency to 8.x should fix this. See https://github.com/testing-library/dom-testing-library/issues/756

I quickly tried this and had to fix some typing errors in dom-selectors.ts. The `Matcher` interface has changed, it’s now

```js
export type MatcherFunction = (
  content: string,
  element: Element | null,
) => boolean
export type Matcher = MatcherFunction | RegExp | number | string
```

This is how the `byTextContent` function could look:

```js
export const byTextContent = (matcher: Matcher, options: MandatorySelectorMatchingOptions): DOMSelector => {
  let textContentMatcher: Matcher;
  const normalizer: NormalizerFn = options?.normalizer || getDefaultNormalizer(options);
  const getTextContent = (elem: Element): string => normalizer(elem.textContent ?? '');

  if (typeof matcher === 'string' || typeof matcher === 'number') {
    textContentMatcher = (_, elem) => {
      if (elem === null) {
        return false;
      }
      if (options?.exact === false) {
        return (
          getTextContent(elem)
            .toLowerCase()
            .indexOf(matcher.toString().toLowerCase()) >= 0
        );
      }

      return getTextContent(elem) === matcher;
    };
  } else if (matcher instanceof RegExp) {
[...]

PR #522, opened 2021-12-19
fix(startup error): update @testing-library/dom to update pretty-format

johncrim:
API change while updating to new @testing-library/dom

fixes: #519

## PR Checklist

## PR Type

```
[x ] Bugfix
[x ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #519

## What is the new behavior?
Karma tests run without failure.

And: Matchers can now be numbers - a number matcher just matches number text, so the behavior is the same as if the matcher were `numberMatcher.toString()`. Because the change to the `Matcher` signature comes along with the dependency update required by this fix, the bug fix and exciting new feature come together.

```
[ x ] Yes
[] No
```

The breaking change will occur due to the change to `MatcherFunction` in `@testing-library/dom` - the MatcherFunction now passes in an `Element`, where it previously was `HTMLElement` - so users will have to test/cast to `HTMLElement` if they need `HTMLElement`-specific properties in their matcher function.

It's unfortunate that this change was added in a minor version of `@testing-library/dom`. Since spectator depends on this external interface that changed, there's no way (that I can find) to fix this major issue without taking the API change.

## Other information

The @testing-library/dom API changed a little, so I had to update `byTextContent()` to fix some typing errors - no issue there, the new API is less strict, so there should be no change to the spectator API.

The issue I found is that type `Matcher` can be a `number`, and if it is a `number` it would have thrown with the previous code. I added a check to prevent that error (also needed to compile), but I'm not sure what the correct behavior is for a `number` matcher, so I'm throwing an exception. If you'd like I can add tests to verify that an exception is thrown, but it's also possible that different behavior is expected if a `number` matcher is passed in. Please clarify what the correct behavior is.

edit: I removed the "throw error on number Matcher" and added support for number matchers in `byTextContent()`.

### DIFF
--- a/projects/spectator/src/lib/dom-selectors.ts
+++ b/projects/spectator/src/lib/dom-selectors.ts
@@ -31,7 +31,7 @@ export const byText: DOMSelectorFactory<SelectorMatcherOptions> = (matcher, opti
 export const byTextContent = (matcher: Matcher, options: MandatorySelectorMatchingOptions): DOMSelector => {
   let textContentMatcher: Matcher;
   const normalizer: NormalizerFn = options?.normalizer || getDefaultNormalizer(options);
-  const getTextContent = (elem: HTMLElement): string => normalizer(elem.textContent ?? '');
+  const getTextContent = (elem: Element | null): string => normalizer(elem?.textContent ?? '');
 
   if (typeof matcher === 'string') {
     textContentMatcher = (_, elem) => {
@@ -47,8 +47,11 @@ export const byTextContent = (matcher: Matcher, options: MandatorySelectorMatchi
     };
   } else if (matcher instanceof RegExp) {
     textContentMatcher = (_, elem) => matcher.test(getTextContent(elem));
-  } else {
+  } else if (typeof matcher === 'function') {
     textContentMatcher = (_, elem) => matcher(getTextContent(elem), elem);
+  } else {
+    // number Matcher not supported
+    throw new Error(`Matcher type not supported: ${typeof matcher}`);
   }
 
   return new DOMSelector(el => DOMQueries.queryAllByText(el, textContentMatcher, options));
--- a/projects/spectator/src/lib/dom-selectors.ts
+++ b/projects/spectator/src/lib/dom-selectors.ts
@@ -33,24 +33,23 @@ export const byTextContent = (matcher: Matcher, options: MandatorySelectorMatchi
   const normalizer: NormalizerFn = options?.normalizer || getDefaultNormalizer(options);
   const getTextContent = (elem: Element | null): string => normalizer(elem?.textContent ?? '');
 
-  if (typeof matcher === 'string') {
+  if (typeof matcher === 'string' || typeof matcher === 'number') {
     textContentMatcher = (_, elem) => {
       if (options?.exact === false) {
         return (
           getTextContent(elem)
             .toLowerCase()
-            .indexOf(matcher.toLowerCase()) >= 0
+            .indexOf(matcher.toString().toLowerCase()) >= 0
         );
       }
 
-      return getTextContent(elem) === matcher;
+      return getTextContent(elem) === matcher.toString();
     };
   } else if (matcher instanceof RegExp) {
     textContentMatcher = (_, elem) => matcher.test(getTextContent(elem));
   } else if (typeof matcher === 'function') {
     textContentMatcher = (_, elem) => matcher(getTextContent(elem), elem);
   } else {
-    // number Matcher not supported
     throw new Error(`Matcher type not supported: ${typeof matcher}`);
   }
 
--- a/projects/spectator/jest/test/dom-selectors/dom-selectors.component.spec.ts
+++ b/projects/spectator/jest/test/dom-selectors/dom-selectors.component.spec.ts
@@ -111,6 +111,20 @@ describe('DomSelectorsComponent', () => {
       });
     });
 
+    describe('with number matcher', () => {
+      it('should match number content', () => {
+        let element = spectator.query(byTextContent(8, { selector: '#number-content-root *' }));
+        expect(element).toHaveId('number-content-only-eight');
+      });
+
+      it('should partially match number with `exact: false`', () => {
+        let elements = spectator.queryAll(byTextContent(8, { selector: '#number-content-root *', exact: false }));
+        expect(elements).toHaveLength(2);
+        expect(elements[0]).toHaveId('number-content-with-eight');
+        expect(elements[1]).toHaveId('number-content-only-eight');
+      });
+    });
+
     describe('with RegExp matcher', () => {
       it('should match the text', () => {
         const element = spectator.query(byTextContent(/^some deeply NESTED TEXT$/, { selector: '#text-content-root' }));
@@ -143,5 +157,6 @@ describe('DomSelectorsComponent', () => {
         expect(element).toHaveId('text-content-span-2');
       });
     });
+
   });
 });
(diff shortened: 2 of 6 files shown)


NetanelBasal:
@benelliott what do you think?

benelliott:
Will take a look

benelliott:
@johncrim These changes look good to me, but while you're there, do you think it would be worth adding support for the `number`-type matcher as per https://github.com/testing-library/dom-testing-library/commit/ec1b642ba6069cbe5a85db649a939ff00d434343# rather than erroring out?

If we keep the code similar to `dom-testing-library` it would be as simple as adding `|| typeof matcher === 'number'` on line 36 and `.toString()` on line 42.

I was trying to decide if this would be a breaking change since we are exposing the type of `Matcher` via our selector APIs. Technically it would since that type has been widened from `HTMLElement` to `Element`, but it's worth noting that this breaking change was actually introduced in a minor release of `dom-testing-library` (v7.29.0).

johnc-ftl:
Thanks @benelliott - that sounds good. I wasn't aware of how the `number` matcher works, so I'll dig into that/follow your recommendations.

RE possible breaking change, in `@testing-library/dom matches.d.ts` it's interesting that `MatcherFunction` was widened to:
```ts
export type MatcherFunction = (
  content: string,
  element: Element | null,
) => boolean
```

But `Match.node` is still HTMLElement:
```ts
export type Match = (
  textToMatch: string,
  node: HTMLElement | null,
  matcher: Matcher,
  options?: MatcherOptions,
) => boolean
```

I haven't dug in enough to determine if this node type is incorrect (would a node be returned of type `Element` but not `HTMLElement` if indicated by the matcher)? If the node type is correct we should be fine - type widening on inputs, but no changes to output shouldn't break anyone. Anyhow this is probably well outside this issue...

benelliott:
Actually due to covariance/contravariance I think the widening of the parameter type of the `MatcherFunction` _could _ break people - e.g. say you have an admittedly weird matcher function `(_, x: HTMLElement) => x.title === 'foo'`, this would be a compilation error now that `x` is `Element` as `title` is defined on the `HTMLElement` subclass.

johnc-ftl:
@benelliott - agreed, I was thinking about it backwards - thanks. 

I think we're in a compatibility pickle then, since existing users have a dependency on @testing-library/dom `Matcher` and `MatcherFunction`, and those signatures have changed. If we update `@testing-library/dom` anything built against those signatures can break (in addition to adding the number matcher, which is non breaking, but should be tested). If we had a spectator-internal interface then we could manage this change.

So, the question is whether to:
1. Update `pretty-format` only, and not take any changes to `@testing-library/dom`. This is the safest change from a compatibility standpoint, but means we won't be able to use anything in newer versions of `@testing-library/dom` without taking the same breaking change.
2. Update `@testing-library/dom` and document the API change in release notes (is there a way to do that?) and risk a compatibility change. 

I'm thinking the best thing to do right now is to kick the can down the road, and update `pretty-format` only - we need that fix urgently. The  `@testing-library/dom` update can be done with a major version update (if anyone expresses a need for it).

Revised PR coming right up...

johnc-ftl:
Crap - I can't figure out a way to update transitive dependencies in a way that is reliable across generally used versions of yarn and npm. Yarn has `resolutions` and npm has `overrides` documented, but they don't seem to be working reliably. I need to update the version in `@testing-library/dom`'s node_modules dir.

So, updating `@testing-library/dom` is the only fix I can find, which comes with the potentially breaking API change. IMO this issue is severe enough (I'm blocked on it) that the API change is worth it.

I will add some tests for the number matcher.

johncrim:
Crap - I can't figure out a way to update transitive dependencies in a way that is reliable across generally used versions of yarn and npm. Yarn has `resolutions` and npm has `overrides` documented, but they don't seem to be working reliably. I need to update the version in `@testing-library/dom`'s node_modules dir.

So, updating `@testing-library/dom` is the only fix I can find, which comes with the potentially breaking API change. IMO this issue is severe enough (I'm blocked on it) that the API change is worth it.

I will add some tests for the number matcher.

johncrim:
Crap - I can't figure out a way to update transitive dependencies in a way that is reliable across generally used versions of yarn and npm. Yarn has `resolutions` and npm has `overrides` documented, but they don't seem to be working reliably. I need to update the version in `@testing-library/dom`'s node_modules dir.

So, updating `@testing-library/dom` is the only fix I can find, which comes with the potentially breaking API change. IMO this issue is severe enough (I'm blocked on it) that the API change is worth it.

I will add some tests for the number matcher.

johncrim:
I removed the "throw error on number Matcher" and added support for number matchers in `byTextContent()`.

Presumably the other methods that take matchers can also take number matchers - this should be handled automatically by `@testing-library/dom`. I didn't add tests for all of them, partly b/c the scope of this has grown a lot from my initial plans - just trying to fix a transitive dependency issue.

I think the breaking change to `MatcherFunction` API is not ideal (I updated the PR info above to describe it), but I think it's necessary to update `@testing-libary/dom`, which is the only fix I can find for this severe/blocking issue. I don't know that it belongs in the docs, but it should be in release notes if available.

NetanelBasal:
I will merge after @benelliott approval. Thanks.

projects/spectator/src/lib/dom-selectors.ts
```diff
@@ -31,24 +31,27 @@ export const byText: DOMSelectorFactory<SelectorMatcherOptions> = (matcher, opti
...
-      return getTextContent(elem) === matcher;
+      return getTextContent(elem) === matcher.toString();
     };
   } else if (matcher instanceof RegExp) {
     textContentMatcher = (_, elem) => matcher.test(getTextContent(elem));
-  } else {
+  } else if (typeof matcher === 'function') {
     textContentMatcher = (_, elem) => matcher(getTextContent(elem), elem);
+  } else {
+    // number Matcher not supported
```
benelliott, line 53: Outdated comment

benelliott (changes requested):
Thanks @johncrim for the effort with this - once you remove that outdated comment I think this looks good!

Not sure if @NetanelBasal has any thoughts on the minor breaking change.

projects/spectator/src/lib/dom-selectors.ts
NetanelBasal: @benelliott I'm not familiar with this library.

NetanelBasal: We could bump to a major release if you guys think this is a breaking change. @benelliott @johncrim

benelliott: It is technically a breaking change due to the change to `MatcherFunction` described above, but it's a fairly small breaking change at that.

NetanelBasal: If it'll break someone app it should be a major release

johncrim: I agree, though it was not a major release for @testing-library/dom. I wish they hadn't made that change.

As long as it's not big deal to you, I think the major update is correct. This change does make a major version change to the @testing-library/dom dependency, which is part of the spectator API.

johncrim:
Thanks @benelliott and @NetanelBasal - comment removed.

benelliott (approved):

benelliott:
Thanks!

NetanelBasal:
@johncrim, Can you please update the commit message to include the breaking change so that the standard version package understands it?

johncrim:
@NetanelBasal I rebased to fix the commit message - also improved the wording of the feat(number matcher) commit, and squashed the last commit. So, same file changes, hopefully improved commit wording.

PR #573, opened 2022-07-25
fix: unpatch `setTimeout` before importing jQuery so it is not captured by zone.js

arturovt:
This is a follow-up of #571.

Alan Agius responded in my issue that:

> You can provide a helper method that does the patching from a secondary entry-point ex: library-with-require/testing.

Initially, `require` worked for me because I used Jest (not Karma).

I've called the secondary entry-point `internals`, but I don't mind changing it to anything else.


NetanelBasal:
@arturovt ci is failing

arturovt:
> @arturovt ci is failing

Forgot to update path aliases

Issue #607, opened 2023-05-15
Template object type safety in createSpyObject


szabyg:
### Description

createSpyObject allows creating mock objects for services providing a template object with mock implementations of the methods of your choice. The only problem is that the type definition of the template parameter is `Partial<Record<keyof T, any>>` allowing to define anything as values.

`export function createSpyObject<T>(type: Type<T>, template?: Partial<Record<keyof T, any>>): SpyObject<T>`

See [/ngneat/spectator/projects/spectator/jest/src/lib/mock.ts#L13](https://github.com/ngneat/spectator/blob/12b7ebe654a67bac56f3cb9997b4cb686a75f311/projects/spectator/jest/src/lib/mock.ts#L13)

Would the definition be `Partial<T>`, the compiler would also check the value types and help to keep the unit test implementation in sync with the code to be tested.

Changing this would cause now many failing tests that were "happy" so far but probably "wrong" in the sense of type safety, and - at least in some cases also - from the logic.

This change would be very helpful when refactoring code.

### Proposed solution

I propose changing the type definition to 

`export function createSpyObject<T>(type: Type<T>, template?: Partial<T>): SpyObject<T>`

I am not sure though how to deal with backward compatibility when making type safety kick in stronger than before.

### Alternatives considered

Not sure. An opt-in solution would be good to allow - especially big projects - to fix their code bit by bit.

### Do you want to create a pull request?

No

szabyg:
I could create a PR but I feel this needs a discussion first.

NetanelBasal:
We can release a new major version.

PR #609, opened 2023-05-16
fix(spectator): make mock template type safe (#607)

szabyg:
BREAKING CHANGE: introducing type safety for the template values in createSpyObject and mockProvider. This might cause breaking builds because the compiler will find issues hidden until now.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[X] Other... Please describe: Stronger type checks for test mock implementation.
```

Weak type checking when mocking services, only checking the template keys, not the values types.

Issue Number: #607 

## What is the new behavior?
The TS compiler will show up mismatches between the service template values and the actual service methods.

```
[X] Yes
[ ] No
```

Introducing type safety for the template values in createSpyObject and mockProvider. This might cause breaking builds
because the compiler will find issues hidden until now.

### DIFF
--- a/projects/spectator/src/lib/mock.ts
+++ b/projects/spectator/src/lib/mock.ts
@@ -55,9 +55,9 @@ export function installProtoMethods<T>(mock: any, proto: any, createSpyFn: Funct
       mock[key] = createSpyFn(key);
     } else if (descriptor.get && !mock.hasOwnProperty(key)) {
       Object.defineProperty(mock, key, {
-        set: value => (mock[`_${key}`] = value),
+        set: (value) => (mock[`_${key}`] = value),
         get: () => mock[`_${key}`],
-        configurable: true
+        configurable: true,
       });
     }
   }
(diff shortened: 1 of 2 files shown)


NetanelBasal:
I'd love to see basic types test. You can use the same library I'm using [here](https://github.com/ngneat/elf/blob/master/packages/store/src/lib/types.spec.ts#L42)

szabyg:
@NetanelBasal Interesting idea and testing library. We spent some hours now to try to find a way to add a meaningful test without testing the type of something that's part of the test itself. On the other hand, the point of the actual change _is_ that the compiler will throw an error if the template type for the provider doesn't match the provider type.

Trying the following things:


```
import { mockProvider } from '@ngneat/spectator';
import { expectTypeOf } from 'expect-type';

class TestProvider {
  testMethod(a: string): string {
    return '';
  }
}
```

This would be nice but it doesn't work:

```
expectTypeOf(mockProvider).parameters.toMatchTypeOf<[TestProvider, { testMethod: (x: string) => string }]>();
expectTypeOf(mockProvider).parameters.not.toMatchTypeOf<[TestProvider, { testMethod: (x: string) => number }]>();
```

Or this
```
expectTypeOf(mockProvider).toBeCallableWith(TestProvider, {
  testMethod: (s: string): number => s.length,
});
```
but `.toBeCallableWith` never fails, no matter what parameter I give it.

Would you have a hint what else I could try? Otherwise I suggest taking it as is, the compiler does complain and that's the point, isn't it?

szabyg:
Any ideas/update on this?
I think type safety is nice so the community would benefit from this change.
@NetanelBasal WDYT?

PR #622, opened 2023-07-06
fix(matchers): use computed styles to determine hidden-ness

aniederer-chatham:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, .toBeHidden cannot catch cases where an element is hidden through CSS, be it a separate style element, constructed stylesheets, or just CSS variables in the `style` attribute. This change should catch all of those cases.

Issue Number: N/A

## What is the new behavior?
Elements hidden with separate style element, constructed stylesheets, or CSS variables are now correctly determined to be hidden

```
[ ] Yes
[x] No
```

This test contains a breaking change in the sense that it can cause tests which were previously passing to fail and vice versa, but those affected tests are likely passing or failing in error. I'll leave that judgment about whether this warrants a major version bump to the maintainers of this library.

### DIFF
--- a/projects/spectator/jest/test/matchers/matchers.spec.ts
+++ b/projects/spectator/jest/test/matchers/matchers.spec.ts
@@ -19,6 +19,7 @@ interface Dummy {
     <div id="classes" class="class-a class-b">Classes</div>
     <div id="styles" style="background-color: indianred; color: chocolate; --primary: var(--black)"></div>
     <custom-element style="visibility: hidden"></custom-element>
+    <div id="computed-style"></div>
   `,
 })
 export class MatchersComponent {}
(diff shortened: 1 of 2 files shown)


projects/spectator/jest/test/matchers/matchers.spec.ts
```diff
@@ -121,6 +122,16 @@ describe('Matchers', () => {
         document.querySelector('custom-element')?.shadowRoot?.querySelector("#shadow-dom")
       ).toBeHidden();
     });
+
+    it('should detect elements whose computed styles are display: none', () => {
+      window.getComputedStyle = () => ({ getPropertyValue: (style) => style == 'display' && 'none' });
```
aniederer-chatham, line 127: I mocked getComputedStyle here instead of testing an actual <style> element or CSS variable case because `window.getComputedStyle` doesn't appear to work properly in this configuration of JSDOM - it will always return `{}` regardless of element. I did test this method manually in my own project, and it works as expected.

NetanelBasal, line 127: What is 'display' && 'none' ?

projects/spectator/jest/test/matchers/matchers.spec.ts
```diff
@@ -121,6 +122,16 @@ describe('Matchers', () => {
...
+      expect(document.querySelector('#computed-style')).toBeHidden();
```
NetanelBasal, line 128: Please check toggle functionality - hidden and than visible

aniederer-chatham: Added in [1c850d1](https://github.com/ngneat/spectator/pull/622/commits/1c850d11bfedad50e40752a6d1a22c6678398b71)

PR #641, opened 2024-02-25
feat(spectator): support defer block behavior

profanis:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/ngneat/spectator/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

### DIFF
--- a/projects/spectator/src/lib/spectator/spectator.ts
+++ b/projects/spectator/src/lib/spectator/spectator.ts
@@ -51,42 +51,46 @@ export class Spectator<C> extends DomSpectator<C> {
     this.detectChanges();
   }
 
-  public get deferBlocks(): DeferBlocks {
-    return this._deferBlocksForGivenFixture(this.fixture.getDeferBlocks());
+  public deferBlock(deferBlockIndex = 0): DeferBlocks {
+    return this._deferBlocksForGivenFixture(deferBlockIndex, this.fixture.getDeferBlocks());
   }
 
   /**
    *
-   * @param deferBlockFixture Defer block fixture
+   * @param deferBlockFixtures Defer block fixture
    * @returns deferBlock object with methods to access the defer blocks
    */
-  private _deferBlocksForGivenFixture(deferBlockFixture: Promise<DeferBlockFixture[]>): DeferBlocks {
+  private _deferBlocksForGivenFixture(deferBlockIndex = 0, deferBlockFixtures: Promise<DeferBlockFixture[]>): DeferBlocks {
     return {
-      renderComplete: async (deferBlockIndex = 0) => {
+      renderComplete: async () => {
         const renderedDeferFixture = await this._renderDeferStateAndGetFixture(
           DeferBlockState.Complete,
           deferBlockIndex,
-          deferBlockFixture
+          deferBlockFixtures
         );
 
         return this._childrenDeferFixtures(renderedDeferFixture);
       },
-      renderPlaceholder: async (deferBlockIndex = 0) => {
+      renderPlaceholder: async () => {
         const renderedDeferFixture = await this._renderDeferStateAndGetFixture(
           DeferBlockState.Placeholder,
           deferBlockIndex,
-          deferBlockFixture
+          deferBlockFixtures
         );
 
         return this._childrenDeferFixtures(renderedDeferFixture);
       },
-      renderLoading: async (deferBlockIndex = 0) => {
-        const renderedDeferFixture = await this._renderDeferStateAndGetFixture(DeferBlockState.Loading, deferBlockIndex, deferBlockFixture);
+      renderLoading: async () => {
+        const renderedDeferFixture = await this._renderDeferStateAndGetFixture(
+          DeferBlockState.Loading,
+          deferBlockIndex,
+          deferBlockFixtures
+        );
 
         return this._childrenDeferFixtures(renderedDeferFixture);
       },
-      renderError: async (deferBlockIndex = 0) => {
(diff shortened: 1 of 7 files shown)


projects/spectator/test/defer-block.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      component: DummyComponent,
+      deferBlockBehavior: DeferBlockBehavior.Manual,
+    });
+
+    it('should render the complete state', async () => {
+      // Arrange
+      const spectator = createComponent();
+
+      // Act
+      const deferFixture = (await spectator.fixture.getDeferBlocks())[0];
```
NetanelBasal, line 66: What do u think about making the API nicer?

```ts
spectator.deferBlocks.renderComplete();
spectator.deferBlocks.renderPlaceholder();
etc..
```

profanis: That's a great idea! 
Thanks for that. I'll work on this :)

profanis: I followed the same API for the nested defer blocks as well.
Please let me know if this is OK and if it is required something else.

projects/spectator/src/lib/spectator/spectator.ts
```diff
@@ -45,4 +45,79 @@ export class Spectator<C> extends DomSpectator<C> {
     // Force cd on the host component for cases such as: https://github.com/ngneat/spectator/issues/539
     this.detectChanges();
   }
+
+  public get deferBlocks(): DeferBlocks {
```
NetanelBasal, line 49: Any reason we need to create a new reference object (i.e the return value of _deferBlocksForGivenFixture) each time we use this getter?

projects/spectator/src/lib/spectator/spectator.ts
```diff
@@ -45,4 +45,79 @@ export class Spectator<C> extends DomSpectator<C> {
...
+  }
+
+  /**
+   *
+   * @param deferBlockFixture Defer block fixture
+   * @returns deferBlock object with methods to access the defer blocks
+   */
+  private _deferBlocksForGivenFixture(deferBlockFixture: Promise<DeferBlockFixture[]>): DeferBlocks {
+    return {
+      renderComplete: async (deferBlockIndex = 0) => {
```
NetanelBasal, line 60: I think it makes more sense to call `spectator.deferBlock(optionalIndex).renderComplete()`, etc.

projects/spectator/test/defer-block.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+      `,
+      standalone: true,
+    })
+    class DummyComponent {
+      isVisible = false;
+    }
+
+    const createComponent = createComponentFactory({
+      component: DummyComponent,
+      deferBlockBehavior: DeferBlockBehavior.Playthrough,
```
NetanelBasal, line 24: What do you say about also support here the state?

```ts
 const createComponent = createComponentFactory({
      component: DummyComponent,
      deferBlockBehavior: DeferBlockBehavior.Playthrough,
      deferBlockState: state,
      deferBlockIndex: optionalIndex
})
```

NetanelBasal (commented):
Great job. I've added some comments, let me know what you think. It's also important to update the README after you done so that users will know about this great feature.

projects/spectator/src/lib/spectator/spectator.ts
profanis: Yes, this makes sense. 
It could be more readable to switch between the states 

```ts
const firstDeferBlock = spectator.deferBlock()
firstDeferBlock.renderComplete()
// assert
firstDeferBlock.renderPlaceholder()
// assert
```

I'll convert the getter to a method

projects/spectator/src/lib/spectator/spectator.ts
profanis: I am not sure if I fully understand you, so I'll do my best answering :) 

My first approach was to use the `deferBlock` both from the parent level and the children ones.
I didn't like, though, the idea of exposing an argument for the sake of internal usage.

```ts
// Parent level
public deferBlock(deferBlockIndex = 0, deferBlockFixtures = this.fixture.getDeferBlocks()): DeferBlocks {
```

// Children
image

To avoid that, I created a private function.

I feel that you were referring to something else though.

projects/spectator/test/defer-block.spec.ts
profanis: I think we can do that here, https://github.com/profanis/spectator/blob/3f63c68440f2dc11cb251e49d8ebdc8a9a44a5e9/projects/spectator/src/lib/spectator/create-factory.ts#L162

```ts
    // e.g.
    spectator.deferBlock(1).renderComplete()

    return spectator;
```

profanis: Hmm, I need to think of an elegant way for this. The `render*` methods return a promise and it won't work nicely in the `create-factory`

NetanelBasal: Ok, we can add it in in a different PR.

projects/spectator/src/lib/spectator/spectator.ts
NetanelBasal: It doesn't matter anymore because we are replacing this API to [this](https://github.com/ngneat/spectator/pull/641/files#r1510292191).

NetanelBasal:
Thanks, great work!

Issue #649, opened 2024-03-19
Spectator is forcing to add a transform explicitly if input() has got a transform function


ashutoshpaul:
Let's say I have a `XyzComponent`. I have to pass `id` in route-params to `XyzComponent`.

Inside `XyzComponent`:
`id = input(undefined, { transform: numberAttribute });`

Inside `xyz.component.spec.ts` file
```
spectator = createComponent({
    props: {
      id: input(5, { transform: numberAttribute })
    },
  });
```
as you can see we have to explicitly add { transform: numberAttribute } because the ngneat/spectator v17 doesn't infer transformFn.

Solution that we came up with:

```
type InferInputSignal<T> = T extends InputSignal<infer K, infer _> ? K : T;
type InferInputSignals<C> = {
    [P in keyof C]+?: InferInputSignal<C[P]>;
};

export interface CustomSpectatorOverrides<C> {
    whenStable?: boolean;
    props?: InferInputSignals<C>;
}
```
all we did was to add `infer _`  in InferInputSignal. And we are overriding `SpectatorOverrides` with `CustomSpectatorOverrides`, with this we are able to achieve:
```
spectator = createComponent({ props: { id: 5 }});
```
this might be a workaround. can you please fix this in appropriate way?

NetanelBasal:
@kfrancois what do you think?

kfrancois:
Good catch! Internally, `InputSignal` is defined as follows:
```ts
export declare interface InputSignal<ReadT> extends InputSignalWithTransform<ReadT, ReadT> {
}
```

Meaning that it extends `InputSignalWithTransform`, so we need to update `InferInputSignal`'s conditional clause to this type instead.

There is one point where I think some feedback is necessary: `InputSignalWithTransform<ReadT, WriteT>` has two generic types:
- `ReadT` is the type that the value will be transformed to
- `WriteT` is the type that we can input into the component


Let's say we have the following component:
```ts
class MyComponent {
    disabled = input(false, { transform: booleanAttribute });
}
```

As `disabled` is of type `InputSignalWithTransform<boolean, unknown>`, we can choose to infer either `boolean` or `unknown`. 

If we decide to infer `TWrite`:
```ts
export type InferInputSignal<T> = T extends InputSignalWithTransform<any, infer K> ? K : T;
export type InferInputSignals<C> = {
    [P in keyof C]+?: InferInputSignal<C[P]>;
};


// { disabled?: unknown }
[...]

PR #651, opened 2024-03-20
feat: 🎸 fix input/output type inference

kfrancois:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/ngneat/spectator/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The following types are not inferred correctly:
- `InputFunction` with transform
- `OutputRef`

Issue Number: https://github.com/ngneat/spectator/issues/649

## What is the new behavior?

Types are properly inferred when using `output()` or when using `input()` with transform set.

```
[X] Yes
[ ] No
```

I believe the required peerDependency should now be Angular 17.3.0 - I didn't update this yet but I believe it's necessary (I have not yet checked the behaviour in a repo running an older version)


projects/spectator/src/lib/types.ts
```diff
@@ -24,7 +24,7 @@ export interface QueryOptions<R> {
   parentSelector?: Type<any> | string;
 }
 
-export type EventEmitterType<P> = P extends EventEmitter<infer T> ? T : never;
+export type OutputType<P> = P extends EventEmitter<infer T> ? T : P extends OutputRef<infer T> ? T : never;
```
NetanelBasal, line 27: This is a breaking change because it requires a specific version of Angular. I don't mind to release a major version but you should include it in the peer dependency and the commit message

kfrancois: Understood! I'll push this later today or tomorrow. Thank you for following up so quickly on this 😄

kfrancois: Added - it's a shame this type got changed in a minor version, since I do believe semantically it's a breaking change in Angular.

Any conditional types that relied on `extends InputSignal` broke. I suppose the Angular team considered this as an internal type (or simply forgot about it/thought it wasn't worth the breaking change).

Issue #663, opened 2024-07-09
Make spectator zoneless-compatible


devfservant:
### Description

Since Angular 18, zoneless applications are available experimentally.
The main advantages are described here: https://angular.dev/guide/experimental/zoneless  
It is possible to configure `TestBed` accordingly to ensure the components under test are zoneless-compatible:

```ts
TestBed.configureTestingModule({
  providers: [provideExperimentalZonelessChangeDetection()]
});
```

The spectator equivalent looks like:

```ts
const createComponent = createComponentFactory({
    component: MyComponent,
    providers: [provideExperimentalZonelessChangeDetection()],
  });
```

But, it does not work, as it uses [`waitForAsync`](https://angular.dev/guide/testing/components-scenarios#async-test-with-waitforasync) under the hood:  
https://github.com/ngneat/spectator/blob/v19.0.0/projects/spectator/src/lib/spectator/create-factory.ts#L127

```
MyComponent › should render
    thrown: "Zone is needed for the waitForAsync() test helper but could not be found. Please make sure that your environment includes zone.js"
```

It would be a huge improvement to see Spectator be zoneless-compatible.

Thanks for all your amazing work

### Proposed solution

--

### Alternatives considered

No alternative considerations.

### Do you want to create a pull request?

No

andrei4ik1997:
@NetanelBasal Hi, any updates on this issue?

NetanelBasal:
Hi, You're welcome to create a PR

devfservant:
Since 19.6.1, it is working well, thank you so much!

PR #703, opened 2025-05-23
fix: remove `waitForAsync` (zoneless support)

arturovt:
In this commit, we remove `waitForAsync` as it is no longer necessary. This rule has also been disabled internally at Google and is no longer in use. Its removal enables zoneless support, since `waitForAsync` depends on zone.js.

We also remove `compileComponents()`, as it is only required when `setClassMetadataAsync` is called—typically representing AOT pre-compiled components outside of TestBed.


rhutchison:
Resolves https://github.com/ngneat/spectator/issues/663

PR #707, opened 2025-06-17
feat: add strong typings to function spies on SpyObject

Alanoll:
## PR Checklist

- [X ] The commit message follows our guidelines: https://github.com/ngneat/spectator/blob/master/CONTRIBUTING.md#commit

## PR Type
```
[ ] Bugfix
[X ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The current version of the library, doesn't pass the argument/return information to the spy objects, so chained calls are reduced to `any` types.

```
spyProvider.methodToSpy.and.returnValue({})
```

The `methodToSpy` has a signature of `(parm: string) => boolean` but when chained like above, `returnValue` doesn't enforce type safety that the value is `boolean` because it's signature is `any`

## What is the new behavior?

Type safety is inferred from the Spied object's type, so `returnValue` now is typed to accept `boolean` instead of `any`

```
[ ] Yes
[ ] No
[X] Maybe?
```
Technically, it's a breaking change in that consumers of the library may start getting compile errors where they haven't before. But if those fakes are correct for their tests, they should be the right types.

### DIFF
--- a/projects/spectator/src/lib/mock.ts
+++ b/projects/spectator/src/lib/mock.ts
@@ -3,10 +3,12 @@ import { FactoryProvider, Type, AbstractType } from '@angular/core';
 
 type Writable<T> = { -readonly [P in keyof T]: T[P] };
 
+export declare type UnknownFunction = (...args: Array<unknown>) => unknown & Function;
+
 /**
  * @publicApi
  */
-export interface CompatibleSpy extends jasmine.Spy {
+export interface CompatibleSpy<F extends UnknownFunction = UnknownFunction> extends jasmine.Spy<(...args: Parameters<F>) => ReturnType<F>> {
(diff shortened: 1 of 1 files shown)


projects/spectator/src/lib/mock.ts
```diff
@@ -3,10 +3,12 @@ import { FactoryProvider, Type, AbstractType } from '@angular/core';
 
 type Writable<T> = { -readonly [P in keyof T]: T[P] };
 
+export declare type UnknownFunction = (...args: Array<unknown>) => unknown & Function;
```
Alanoll, line 6: The original checks used `Function` which is why I included it in the type definition, but I'm not sure if it's needed.

